PR #28, opened 2026-06-26
security: harden CRON_SECRET guard — fail-closed + timing-safe comparison

skizzy1986:
Two bugs in the vitals-import shared-secret guard:
- Fail-open when CRON_SECRET env var is unset (any caller could trigger
  a service-role import run with no credentials)
- Plain !== string comparison is a timing oracle

Fix: fail with 500 when CRON_SECRET is not configured; use
timingSafeEqual from jsr:@std/crypto/timing-safe-equal for the
header comparison.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_0159hNfnbPzUhhJhcNtVvhBV

### DIFF
--- a/supabase/functions/vitals-import/index.ts
+++ b/supabase/functions/vitals-import/index.ts
@@ -41,8 +41,10 @@ Deno.serve(async (req: Request) => {
   if (missing.length) return json({ error: "missing env", missing }, 500);
 
   const fetchCsv = async (url: string, label: string): Promise<string> => {
-    const res = await fetch(url, { redirect: "follow" });
+    const res = await fetch(url, { redirect: "follow", signal: AbortSignal.timeout(10_000) });
     if (!res.ok) throw new Error(`${label} fetch failed: HTTP ${res.status}`);
+    const ct = res.headers.get("content-type") ?? "";
+    if (!ct.startsWith("text/")) throw new Error(`${label} unexpected content-type: ${ct}`);
     return res.text();
   };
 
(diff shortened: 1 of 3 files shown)


skizzy1986:
**Triage verdict (2026-07-02 sweep): still needed — recommend rebase + merge as a P1 security item.**

Verified against current `main`:
- `supabase/functions/vitals-import/index.ts:22-25` is **still fail-open** — `if (expected && …)` means an unset `CRON_SECRET` disables the guard entirely on a service-role function — and still uses a plain `!==` comparison (timing oracle).
- The newer `vitals-import-api/index.ts:34-37` (added after this PR) has the **same fail-open guard shape and timing-unsafe comparison**. It happens to fail-closed in practice because `CRON_SECRET` is also in its missing-env check at line 54, but the comparison is still `!==`.

Suggested path: rebase onto `main` and extend the same fix (fail-closed + `timingSafeEqual`) to `vitals-import-api`, or redo as one fresh PR covering both functions. Tracked in the Sprint 1 triage issue.

---

skizzy1986:
Closing — despite the earlier verdict recommending rebase+merge, issue #113's sprint tracking confirms the fix shipped separately in #131 (merged 2026-07-02): fail-closed + timingSafeEqual applied to both `vitals-import` and `vitals-import-api`, i.e. broader scope than this PR covered.

---

Issue #79, opened 2026-06-29
refactor: relocate StrengthLogger.jsx into views/ + tests (next month)


skizzy1986:
**Overflow from the month plan — completes the rest of #53 alongside the burndown.**

`web/src/StrengthLogger.jsx` (1,665 lines, untested, root-level) → relocate under `views/` and add React Testing Library coverage (render + core interactions). Also pick up the mobile-view tests called out in #53 (`MobileApp`, `MobileRecovery`, `MobileStrength`, `MobileSessionLog`). Remove each from the C1 coverage `exclude` as it gains tests. Part of #52; closes the testing gaps in #53.

Issue #61, opened 2026-06-29
CI/CD hardening — raise the pipeline to a high standard before the refactor leans on it


skizzy1986:
Tracking issue for the **C-series** CI/CD hardening work. The current pipeline is already decent (concurrency cancellation, npm caching, `npm audit` gate, coverage PR comments, build-gated-on-test). This issue closes the real gaps so quality is **enforced**, not hoped-for — and lands the standard **before** the monolith refactor (#52) leans on it.

## Why now
Coverage scope is **implicit**: `web/vite.config.js` has no `coverage.include`/`exclude`, so v8 only counts files a test happens to import. The 8,715-line monolith is never imported by a test → it is **invisible to the 70/70/60 gate**, so coverage currently passes *by accident*. Making scope explicit converts coverage into an honest **ratchet** that enforces "test every extraction" during #52.

## Baseline bundle
- [ ] **C1 — Coverage scope + ratchet (keystone, must land before any extraction)**
- [ ] **C2 — Workflow hygiene bundle** (conventional-commit PR-title lint, dependency-review, SHA-pin/standardize action versions, auto-merge-on-green)
- [ ] **C3 — Pre-push test hook + CodeQL SAST**

## Heavier add-ons (all four chosen)
- [ ] **C4 — E2E smoke tests** (Playwright; assert each tab renders)
- [ ] **C5 — Bundle-size budget + Lighthouse/a11y CI**

Each item is one PR. See the per-PR sub-issues linked below. Stale-doc note: real gates are 70/70/60 (`vite.config.js:55-59`) but `CLAUDE.md` says 50/30/35 — reconciled in the #52 H0 doc PR.

skizzy1986:
**Closing as completed** (2026-07-02 backlog sweep) — all five C-series items shipped:

- [x] **C1 — Coverage scope + ratchet** → merged as **#81** (explicit `include`/`exclude` in `web/vite.config.js`; ratchet currently at 52/50/48 with per-file 80/80/70 gates on extracted files)
- [x] **C2 — Workflow hygiene bundle** → PR-title lint **#84**, dependency-review **#87**, auto-merge enabled; SHA-pinning reframed and tracked separately in **#63**
- [x] **C3 — Pre-push test hook + CodeQL** → merged as **#83** (`.husky/pre-push`, `codeql.yml`)
- [x] **C4 — Playwright E2E smoke tests** → merged as **#88** (`e2e-web.yml`, currently informational)
[...]

Issue #52, opened 2026-06-29
Refactor burndown — resume strangler-fig decomposition of the monolith


skizzy1986:
Tracking issue for the stalled strangler-fig refactor. The monolith is the main barrier to extending the app, and because the refactor isn't a backlog item it never competes for time against features.

## Current state (2026-06-29)
- `web/src/erg-dashboard.jsx` — **9,733 lines** (the primary monolith)
- `web/src/StrengthLogger.jsx` — **1,665 lines** (unextracted, co-located at root)
- `web/src/main.jsx` — large auth entry point; **no `App.jsx` yet**
- Extracted so far: `constants/`, `hooks/`, `utils/`, some `components/` + `views/` (~4.3k lines modular). Net progress ~1,000 lines in ~5 months.

## Approach
Follow the strangler-fig order in `.claude/skills/architecture.md`: constants → utils → hooks → components → views → rename entry to `App.jsx`. **One extraction per PR**, app stays buildable after each. Use `/refactor <module>`.

## Burndown checklist (extract one per PR; check off as merged)
- [ ] Identify and extract the next view from `erg-dashboard.jsx`
- [ ] Extract `StrengthLogger.jsx` into `components/`/`views/` with tests
- [ ] Continue view-by-view extraction (track line count per PR in this issue)
- [ ] Reach `App.jsx` milestone (entry point renamed, monolith dissolved)

Update the line counts in this issue after each merge so the burndown is visible.

skizzy1986:
**Burndown update — refactor + commercial-baselines sprint (2026-07-01)**

Two more tabs extracted and merged this sprint, each tested to the commercial **80/80/70** per-file gate:

| PR | Tab → view | Monolith after |
|---|---|---|
| #106 | `strength` → `views/StrengthView.jsx` | 6,303 |
| #107 | `mobility` → `views/MobilityView.jsx` (+ new `constants/mobility.js`) | **5,834** |

`web/src/erg-dashboard.jsx` is now **5,834 lines** (the header's 9,733 was pre-#89; the schedule/analysis/Journal/Recovery extractions had already taken it to ~6,581 before this sprint began).

Also landed alongside: runtime error monitoring (Sentry, #105) and the commercial coverage destination (80/80/70, ratcheting per-file as each view lands).

[...]

PR #81, opened 2026-06-29
ci: make coverage scope explicit + ratchet from honest baseline

### LINKED ISSUE #62: ci: make coverage scope explicit + ratchet (keystone)
**C1 — the keystone. Must merge before any extraction PR under #52.**

## Problem
`web/vite.config.js` sets thresholds (70/70/60) but no `coverage.include`/`exclude`. v8 only counts files imported by a test, so the untested 8,715-line monolith is invisible to the gate — coverage passes by accident.

## Change
- Add `coverage.include: ['src/**']` and `all: true` so untested-but-in-scope files count.
- Add `coverage.exclude` for the not-yet-extracted monolith + entries: `erg-dashboard.jsx`, `StrengthLogger.jsx`, `main.jsx`, `test-setup.js`, plus pure-data `constants/**` if appropriate.
- Set thresholds to the **current real baseline** measured after scoping, and ratchet upward as coverage rises.

## The discipline this unlocks
Every later extraction PR removes exactly the file(s) it extracts from the `exclude` list **in the same PR** — so a file cannot leave the exclude list without hitting 70%. This makes "test every extraction" mechanically enforced rather than a convention.

## Acceptance
- `npx vitest run --coverage` reports against an explicit, intentional file set.
- The monolith is excluded (documented inline why), and the gate is green at the measured baseline.
- A short note in the PR body records the starting numbers for the ratchet.

skizzy1986:
Closes #62

## What changed and why

Makes the coverage gate honest: v8 only counts files a test happens to import, so with no `coverage.include`/`all` the 8,715-line monolith was invisible to the gate and coverage was passing **by accident** (~74% measured across only ~17 imported files). This adds `coverage.all` + `include: ['src/**']` so every in-scope source file counts, excludes the not-yet-extracted surfaces, and resets the thresholds to the real measured baseline so the gate becomes an honest **ratchet**.

**Scope now explicit (`web/vite.config.js`):**
- `all: true` + `include: ['src/**']` — count every in-scope file, not just test-imported ones.
- `exclude` (each documented inline): `erg-dashboard.jsx` (the monolith, #52), `StrengthLogger.jsx` (#79), `main.jsx` (auth/bootstrap entry), `constants/**` (pure data), `test-setup.js` (the harness).

**Baseline measured after scoping** — the honest starting numbers:

| Metric | Before (by accident) | After (real) | Threshold set |
|---|---|---|---|
| Lines | 74.74% | **48.98%** | 48 |
| Functions | 71.34% | **46.81%** | 46 |
| Branches | 65.46% | **40.38%** | 40 |

**The discipline this unlocks:** every later extraction PR under #52 removes the file(s) it extracts from `exclude` **in the same PR**, so a file cannot leave the exclude list without meeting threshold — "test every extraction" is now mechanically enforced. Ratchet the thresholds up as coverage rises; never lower them.

Keystone of the C-series (#61). **Must merge before any extraction PR under #52.**

## How to test manually

Not needed — config-only change, fully exercised by CI. To verify locally: `cd web && npx vitest run --coverage` reports against the explicit file set and exits 0 at the baseline above.

## Checklist

- [x] CI is green (lint, test, build) — verified locally: coverage exit 0, build exit 0, lint 0 errors, format clean
- [x] Tests cover the change, or I've noted why they don't — config-only; the existing 193-test suite is the exercise

---

> **Heads-up (out of scope, not fixed here):** `web/package-lock.json` has no entry for `@vitest/coverage-v8` though it's in `package.json` devDeps — CI's `npx vitest run --coverage` auto-fetches it at runtime so coverage works, but `npm ci` won't install it. Worth a follow-up `npm install` + lockfile commit to pin it. Kept out of this keystone to stay minimal.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MLks4YNjeoUsUHc5SDLVkx)_

### DIFF
--- a/web/vite.config.js
+++ b/web/vite.config.js
@@ -52,10 +52,36 @@ export default defineConfig({
     coverage: {
       provider: 'v8',
       reporter: ['text', 'json-summary', 'json'],
+      // Make scope explicit. Without `include` + `all`, v8 only counts files a
+      // test happens to import, so the untested monolith is invisible and the
+      // gate passes by accident. `all: true` counts every in-scope src file.
+      all: true,
+      include: ['src/**'],
+      // Excluded from the coverage denominator (documented why each is here):
+      //  - erg-dashboard.jsx: the not-yet-decomposed monolith (#52). Each
+      //    extraction PR removes the file(s) it extracts from this list in the
+      //    same PR, so a file cannot leave `exclude` without hitting threshold.
+      //  - StrengthLogger.jsx: large untested component, extraction tracked (#79).
+      //  - main.jsx: auth/bootstrap entry point, not unit-testable in jsdom.
+      //  - constants/**: pure data tables — no logic to cover; counting them
+      //    only distorts the denominator.
+      //  - test-setup.js: the test harness itself.
+      exclude: [
+        'src/erg-dashboard.jsx',
+        'src/StrengthLogger.jsx',
+        'src/main.jsx',
+        'src/constants/**',
+        'src/test-setup.js',
+      ],
+      // Baseline measured 2026-06-29 after making scope explicit (was passing
+      // by accident at ~74% because only test-imported files counted). These
+      // are the honest starting numbers — RATCHET THEM UP as each extraction
+      // PR removes a file from `exclude` and lands its tests. Never lower them.
+      //   measured: lines 48.98 / functions 46.81 / branches 40.38
       thresholds: {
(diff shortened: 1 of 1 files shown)


skizzy1986:
Internal change — no release note. This PR updates CI tooling only (v8 coverage scope + threshold ratchet); no user-visible behaviour changed.

---

PR #94, opened 2026-07-01
feat(db): context store — coach_log + anchors tables with RLS

skizzy1986:
## What changed and why

Stands up a Supabase-resident **context store** so project context lives in one backend that both tools read — Coach (chat via MCP) and Code (via DB) — instead of being split across Coach memory, `CLAUDE.md`, and Drive.

Two tables with distinct jobs (kept separate):

| Table | Job | Key detail |
|---|---|---|
| `coach_log` | Append-only diary + decision record | `supersedes` nullable self-FK — reversals write a **new** row, history is never mutated |
| `anchors` | Current calibration + phase state | Partial unique index `(user_id, key) where superseded_at is null` makes "current value per key" a one-row read |

Both tables: RLS enabled, single `ALL` owner policy `user_id = auth.uid()`, `user_id uuid not null default auth.uid()`, native `date`/`timestamptz` throughout (not the legacy `sessions.date` text pattern) — matching the existing modern convention.

Applied live as migration `context_store_coach_log_and_anchors` and seeded with the five structural anchors (`rowing_cp`, `bike_ftp`, `current_phase`, `current_block`, `doctrine_sha` → `0013c9f`) via explicit `user_id`. `coach_log` content is Coach's lane and is intentionally left empty. This file (`006_context_store.sql`) is the version-controlled schema mirror.

## How to test manually

Schema/data change, not exercised by CI. Verified live against project `swdrueaserjzhuxnzmeu`:

```sql
-- current value per key (one live row each): returns 5 rows with exact seeded values
select key, value, unit, status from public.anchors where superseded_at is null order by key;
```
- `coach_log` `supersedes` self-join resolves (checked with throwaway rows, then deleted).
- `get_advisors(security)` shows no findings for either new table (RLS + policy present).

## Checklist

- [x] Tests cover the change, or I've noted why they don't — SQL migration, no JS surface; full suite (290 tests) still passes

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gta8vvNbtZXkFZbM2WhpdL)_

### DIFF
--- a/supabase/migrations/006_context_store.sql
+++ b/supabase/migrations/006_context_store.sql
@@ -1,0 +1,57 @@
+-- 006_context_store.sql
+-- Context store: one Supabase-resident source of truth that both tools read
+-- (Coach chat via MCP, Code via DB). Applied live as migration
+-- `context_store_coach_log_and_anchors`.
+--
+-- Two tables, distinct jobs:
+--   coach_log — append-only diary + decision record (never edited in place;
+--               reversals write a NEW row with `supersedes` -> the old row).
+--   anchors   — current calibration + phase state; exactly one live row per key
+--               (superseded_at IS NULL) so "current value per key" is a one-row read.
+--
+-- Structural seed rows (rowing_cp, bike_ftp, current_phase, current_block,
+-- doctrine_sha) are inserted separately with an explicit user_id — auth.uid()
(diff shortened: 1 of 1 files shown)


skizzy1986:
Internal change — no release note

---

Issue #113, opened 2026-07-02
Sprint 1 — Truth & Triage: owner actions + salvage fixes


skizzy1986:
Tracking issue for the hygiene sprint (analysis 2026-07-02). The automated parts ran today: 4 superseded PRs closed (#91, #48, #44, #33), delivered tracking issues closed (#95, #61), burndown numbers refreshed on #52, doc-reconciliation + Dependabot-fix PR opened from `claude/codebase-sprint-planning-pbibbf`. What remains is **owner-only decisions**.

## Owner actions (Scott)
- [ ] **Merge decisions on the surviving PRs** (triage verdicts commented on each):
  - #111 — recommend merge (clean, fresh, carries a live `ReferenceError` fix)
  - #80 — recommend merge (inert until configured)
  - #28 / #26 / #93 — recommend close-and-redo (see salvage items below)
- [ ] **Install the Claude GitHub App + add `ANTHROPIC_API_KEY` secret** — activates #80's `@claude` automation
- [ ] **Set `ADD_TO_PROJECT_PAT` secret** — `add-to-project.yml` is currently a silent no-op, so nothing reaches the "Split IQ" board automatically
- [ ] **One-time SHA-pin pass** from a networked machine (`pin-github-action` / `ratchet` / `frizbee`) — completes #63; Dependabot maintains pins afterwards
- [ ] **Strava paid-tier decision** — #54 is research-complete but blocked on the API subscription

## Salvage fixes (small fresh PRs; each replaces an unrebaseable stale PR)
- [x] **CRON_SECRET guard hardening** (replaces #28, P1 security) — shipped in #131 (merged 2026-07-02): fail-closed + `timingSafeEqual` (via SHA-256 digest compare) applied to both `vitals-import` and `vitals-import-api`.
- [x] **TSS/mobile data fixes** (replaces #26) — shipped in #132: root cause was `useTSSHistory.js` filtering on a `status` value (`'logged'`) that never appears in production data, so the hook silently returned empty 100% of the time (not just on same-day collisions as originally scoped) and every consumer fell back to a stale seed array. Fixed status filter, added a duration-text parser, added date normalization, fixed same-day TSS aggregation, and switched `MobileAnalytics` RECENT SESSIONS to live data. Follow-ups filed: #133 (pre-existing UPCOMING-section date bug) and #134 (CLAUDE.md status-value doc drift).

## Exit criteria
[...]

PR #26, opened 2026-06-25
fix(mobile): live TSS aggregation, live recent-sessions list, mobile-view tests

skizzy1986:
Closes the two residual gaps in P2 (live-data analytics) that landed in `main` via PR #25, plus the missing mobile-view test coverage the cloud handover flagged.

## Changes
- **Gap A — TSS per-day aggregation (`useTSSHistory`)**: previously emitted one `{date, tss}` row per session. `calcTrainingLoad` keys its daily map by date, so two logged sessions on the same day silently dropped one. Now sums `duration*srpe/60` per date (raw, then rounds once) so same-day sessions combine correctly.
- **Gap B — live "recent sessions" list (`MobileAnalytics`)**: the RECENT SESSIONS list rendered the hardcoded `DAILY_TSS` seed. Now driven by live `useSessions` (`status !== 'planned'`, newest 5), falling back to the seed when the DB is empty (new install / offline).
- **Tests**: RTL tests for `MobileSessionLog`, `MobileStrength`, `MobileRecovery` (loading / empty / populated / interaction paths) and the `MobileAnalytics` recent-sessions live-vs-seed logic. Added a `ResizeObserver` stub to `test-setup.js` so recharts' `ResponsiveContainer` renders under jsdom.

## Verification
- `npm run lint` — 0 errors (2 pre-existing `usePM5.js` warnings, unrelated)
- `npm test` — **101 passing** (was 84; +2 aggregation, +15 mobile-view)
- `npm run build` — exits 0

No changes to `erg-dashboard.jsx` (desktop monolith) — deferred to avoid conflict with the parallel strangler-fig refactor.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #132, opened 2026-07-02
fix: TSS/CTL/ATL pipeline reads real session data instead of a stale seed

skizzy1986:
Closes the TSS/mobile item in #113 (a sub-item — the rest of #113 tracks unrelated owner actions and stays open). Supersedes stale PR #26.

## What changed and why

`useTSSHistory.js` filtered on `status = 'logged'`, but production `sessions` rows never use that value for historical data — the real statuses are `actual`/`completed` (bulk-imported history, 44 rows) and `logged` (the live PM5 Bluetooth save path in `ErgLiveView.jsx`, confirmed by reading its insert call directly). The hook was silently returning an **empty array in production 100% of the time**, not just on same-day collisions as originally reported — every consumer (`MobileAnalytics`, `MobileRecovery`, `useCoach`) was falling back to a hardcoded 6-week-old seed array (`DAILY_TSS`) the whole time.

Root-cause fixes:
- `useTSSHistory`/`MobileAnalytics` now match all three real "done" statuses via a new shared `constants/sessionStatus.js` (`actual`, `completed`, `logged`) — a mid-review fix after the first pass only included `actual`/`completed` and would have silently excluded newly-logged live PM5 sessions going forward.
- New `parseDurationMinutes()` parses the free-text `duration` column's real production formats (`"45:00"`, `"57m"`, `"1h4m"`, `"2h00m"`, bare numbers) — the old code multiplied the raw string directly, producing `NaN` for nearly every real row.
- New `toISODate()` normalizes the `M/D/YY` (not zero-padded) date text before it's used as a map key or calendar-range bound in `calcTrainingLoad`, so lookups and chronological ordering are no longer silently wrong.
- `calcTrainingLoad` now sums (not overwrites) TSS for sessions sharing a date, and concatenates their notes.
- `MobileAnalytics` RECENT SESSIONS switches from the hardcoded seed to live session data, using the same live-preferred/seed-fallback precedence already used elsewhere in the component.

## How to test manually

Not required — covered by the new/updated automated tests below. If you want to sanity-check visually: open the mobile Analytics tab and confirm the CTL/ATL/TSB chart and RECENT SESSIONS list reflect actual logged sessions rather than dates in the 2026-05-22 to 2026-06-13 range (the seed's fixed date range).

## Checklist

- [x] CI is green (lint, test, build) — `npm run build`, `npm run lint` (0 errors), `npm run format:check`, `npm test` (367/367) all pass locally.
- [x] Tests cover the change — new `duration.test.js`, `dateFormat.test.js`, `MobileAnalytics.test.jsx` (previously zero coverage), `MobileRecovery.test.jsx` (spot-check, no source change), plus additive cases in `useTSSHistory.test.js` and `trainingLoad.test.js` (including a same-day 3+ entries case and a case proving `status: 'logged'` sessions are now included).
- [x] No unexpected console errors in the browser — not verified in a live browser session in this sandbox; covered by RTL component tests instead.

## Follow-ups filed separately (found during review, correctly out of scope for this PR)
- A pre-existing bug in `MobileAnalytics.jsx`'s UPCOMING section: it compares raw `M/D/YY` session dates against an ISO `today` string, which can hide genuinely-future sessions and show genuinely-past ones.
- `CLAUDE.md`'s documented `sessions.status` values (`"logged"`/`"planned"` only) are stale relative to production reality (`actual`/`completed`/`planned`/`cancelled`/`logged`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xaz1QVoRQ9zaue2eubjNQQ)_

### DIFF
--- a/supabase/functions/vitals-import-api/cronGuard.ts
+++ b/supabase/functions/vitals-import-api/cronGuard.ts
@@ -1,0 +1,42 @@
+// cronGuard.ts — fail-closed, timing-safe shared-secret check for the
+// x-cron-secret header. Call this before any other work happens.
+//
+// Threat model this closes:
+//   1. CRON_SECRET unset or "" must NEVER authorize a request (previously:
+//      `if (expected && ...)` skipped the check entirely when `expected`
+//      was falsy, silently opening the endpoint to any unauthenticated
+//      caller).
+//   2. Raw string `!==` short-circuits on the first differing byte, so a
+//      naive check leaks how many leading characters of a guessed secret
+//      are correct via response timing. We instead compare fixed-length
+//      SHA-256 digests of both sides with node:crypto's timingSafeEqual,
+//      so every call — right or wrong, short or long, matching or not —
+//      does the identical fixed-size constant-time compare. Hashing first
+//      also sidesteps timingSafeEqual's hard requirement that both inputs
+//      be equal length (it throws otherwise): the two digests are always
+//      exactly 32 bytes, so there is no length branch to leak and no
+//      exception path to reason about.
+import { createHash, timingSafeEqual } from "node:crypto";
+
+const unauthorized = (): Response =>
+  new Response(JSON.stringify({ error: "unauthorized" }), {
+    status: 401,
+    headers: { "Content-Type": "application/json" },
+  });
+
+const sha256 = (input: string): Uint8Array => createHash("sha256").update(input, "utf8").digest();
+
+/**
+ * Returns a 401 Response if the request fails the CRON_SECRET guard.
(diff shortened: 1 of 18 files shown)


skizzy1986:
**Release note (Fixed)**

Fixed: the Analytics tab's fitness chart (CTL/ATL/TSB) and Recent Sessions list were silently showing 6-week-old placeholder data instead of your real training history — a status-matching bug meant every logged session was being filtered out. Both now reflect your actual logged sessions, and session duration/date parsing (e.g. `"1h4m"`, `"57m"`) is fixed so training load is calculated correctly, including days with multiple sessions.

---

PR #154, opened 2026-07-17
feat: migrate shared components to THEME tokens (design tokens slice 2b)

skizzy1986:
## What

Second sub-slice of the inline-literal migration (follow-up to #148, builds on #153). Replaces ~60 inline hex literals with `THEME` token references across the 11 shared components:

LogEntry, WorkoutItem, LoadTooltip, LogSessionForm, ErgTooltip, StrengthTooltip, LiveMetric, WorkoutTarget, PaceTrendChart, mobile/BottomTabBar, ErrorFallback.

**Conventions applied (per the approved spec):**
- Alpha-suffixed literals → `` `${THEME.token}NN` `` template literals (8-hex shape preserved, never rgba()).
- Runtime-variable alpha bases (`${color}50`) untouched.
- Recharts/SVG bright-line: 8 attribute colors in PaceTrendChart deferred to the Recharts slice; only `style={{}}` colors migrated.
- Intentional survivors: `#ffaa44` (LogEntry coach-note) and `#7a9a8a` (WorkoutItem done-state) — no matching token; flagged for a future token decision rather than silently invented.

**Rendered-color changes (exactly two, both pre-approved drift folds):**
- `#0a0a0f` → `THEME.bg` `#08080d` (ErrorFallback ×2)
- `#a0a0b8` → `THEME.muted` `#7e7e9a` (WorkoutTarget)

All other swaps are visually identical (incl. `#888`/`#fff` shorthand → 6-digit expansion).

## Pipeline evidence

Factory chain build → verify → review + validate; Gate 3 approved by Scott.
- Tests: 43 files / 386 tests pass, exit 0. (One 5s timeout observed under concurrent local runs reproduced-away in isolation — documented Drive-checkout flake.)
- Coverage: verified **digit-for-digit identical** pre/post diff via stash-differential coverage runs; per-file 80/80/70 gates (ErrorFallback, WorkoutItem, LoadTooltip, LogEntry) all clear.
- Lint 0 errors; build exit 0.
- Judges: Code Reviewer **APPROVE**; Reality Checker **ship** (no Blockers/Should-fixes); Analyzer all criteria **MET**.

## Next

PR 2c — desktop views (~500 literals: OverviewView, ErgView, RecoveryView, JournalView, MobilityView, StrengthView, CalendarView, LogView, PlanView, ErgLiveView).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #151, opened 2026-07-14
refactor: extract useSessionLog hook from App.jsx

skizzy1986:
## What

"PR B" of the App Rename Sprint — the deferred half of PR #145's scope. Verbatim strangler-fig extraction of the Supabase sessions data block out of `App.jsx` into a hook.

## Changes

- **New `web/src/hooks/useSessionLog.js`** (86 lines) — moves, byte-identical: the retired `sessionLog` seed, the `dbSessions`/`dbStatus` state, `fetchSessions` (fetch + row mapping via `normType`), the mount effect, the db-first seed merge, the planned/logged split, and `loggedKeys`. Returns `{ dbStatus, allSessions, loggedSessions, plannedSessions, loggedKeys, fetchSessions }`.
- **`web/src/App.jsx`** — 515 → 447 lines; the block is replaced by one destructure (only the four names App still uses, keeping `no-unused-vars` green); unused `supabase`/`normType` imports removed. Zero other lines touched.
- **New `web/src/hooks/__tests__/useSessionLog.test.js`** — 4 tests: mapping/success path, error fallback, planned/logged split + `loggedKeys`, refetch.

### Why `useSessionLog`, not `useSessions`

`hooks/useSessions.js` already exists — the mobile React Query hook (PR #25) with three live consumers (`useCoach.js`, `MobileAnalytics.jsx`, `MobileSessionLog.jsx`). It and its consumers are untouched. A React Query conversion of the extracted block was explicitly deferred; this PR moves code verbatim.

## Verification

- `npm run build` PASS · `npm test` PASS (44 files, 390 tests) · `eslint src/` 0 errors · prettier clean on changed files.
- Coverage ratchet (#147 raises the floor to 68/67/64) is unaffected: the new hook ships with its tests.

## Noted, out of scope

The ESLint flat config (`web/eslint.config.js`) has no `files` pattern covering `.jsx`, so `eslint src/` doesn't actually lint any `.jsx` file — pre-existing gap, worth its own issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #147, opened 2026-07-11
config: ratchet coverage floor to 68/67/64 after App.jsx rename

skizzy1986:
## What

"PR D" of the App Rename Sprint — the coverage ratchet that follows PR #145.

With `App.jsx` (the former monolith) now inside the coverage denominator and carrying its render test, the measured aggregate on `main` is:

| Metric | Measured (2026-07-11) | Old floor | New floor |
|---|---|---|---|
| Lines | 71.76% | 52 | **68** |
| Functions | 71.52% | 50 | **67** |
| Branches | 68.62% | 48 | **64** |

~4 points of headroom is kept for denominator drift as future PRs land. Per the ratchet rule in `web/vite.config.js`, the numbers only go up.

## Verification

- `npx vitest run --coverage` passes locally against the new thresholds (379/379 tests, summary above).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #146, opened 2026-07-11
docs: sweep stale erg-dashboard.jsx references after App.jsx rename

skizzy1986:
## What

Doc-only sweep of stale `erg-dashboard.jsx` references after the entry-point rename landed (PR #145, closing #52). This is "PR C" of the App Rename Sprint.

## Changes

- **CLAUDE.md** — project-structure tree entry and the Strangler Fig section now describe `App.jsx` (~515-line shell/router) and mark the decomposition complete; remaining large files updated to current sizes (#77, #79).
- **README.md** — architecture prose and project-structure tree updated.
- **coach/PROJECT_MANAGEMENT_ANALYSIS.md** — dated addendum (2026-07-11) recording that Finding 5 is fully resolved; historical figures left untouched.
- **.claude/AGENTS.md**, **.claude/skills/{architecture,erg-context,refactor}** — extraction targets and layer maps now name `App.jsx`.

Remaining `erg-dashboard.jsx` mentions are intentionally historical ("formerly erg-dashboard.jsx", the dated PM-analysis findings).

## Verification

- `rg 'erg-dashboard\.jsx'` — only historical/"formerly" mentions remain.
- No code touched; full test suite passed on pre-push hook (379/379).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #171, opened 2026-08-18
fix(ci): break the npm-audit deadlock and pin Playwright to the lockfile

skizzy1986:
## What changed and why

The required `Lint & Format` job ended with an unscoped `npm audit --audit-level=high`. Six advisories had accumulated (critical `tar`, plus `undici`, `brace-expansion`, `fast-uri`, `nanoid`, `postcss`), so the gate failed **every** PR touching `web/**` — including the Dependabot PRs carrying the fixes, because no single bump could clear it while the other five remained. Nothing has merged into `web/` since #156 on 2026-07-20, and four CI-green PRs (#146, #147, #151, #154) have been stranded for a month.

Three changes:

1. **`npm audit fix`** (no `--force`) clears all six. Lockfile only — `package.json` is untouched. This supersedes Dependabot #159, #163, #166, #167, #170.
2. **Scope the gate to `--omit=dev`**, so dev-only transitive advisories (eslint/vitest/workbox tooling) cannot block application PRs again. Anything reaching users is still gated.
   > Worth recording: `--omit=dev` **alone would not have unblocked us**. `tar` and `postcss` are in the production tree, so scoping still reported 5 vulnerabilities. The lockfile fix was required regardless.
3. **Derive the Playwright container tag from the lockfile.** The tag had to be bumped in lockstep with `@playwright/test` and nothing enforced it, so every Dependabot bump of that package killed the entire smoke suite on a missing browser binary — see #168. It resolves to `v1.61.1` today, identical to the previous hardcoded value, so this is behaviour-preserving.

## How to test manually

Not needed — the change is entirely CI config plus a lockfile bump, and is covered by the existing gates. The meaningful signal is **Lint & Format** and **Playwright Smoke** going green on this PR.

## Verified locally

- `npm run lint` — 0 errors (5 pre-existing warnings)
- `npm audit --omit=dev --audit-level=high` — 0 vulnerabilities
- `npx vitest run --coverage` — 386/386 pass; coverage 71.8 lines / 71.7 functions / 68.6 branches
- `npm run build` — succeeds

`npm run format:check` flags 110 files locally, but that is a pre-existing Windows line-ending artefact, not a real failure: `.prettierrc` has no `endOfLine` key (so it defaults to `lf`) while `core.autocrlf=true` produces a CRLF working tree. An LF-normalised copy of an untouched file passes cleanly, and CI checks out LF — which is why this job has passed on every previous PR. Deliberately not running `prettier --write`, as that would rewrite 110 files this PR has no business touching.

## Follow-ups (not in this PR)

- Close the superseded Dependabot PRs and update #157.
- Rebase and land #146 → #147 → #151 → #154.

## Checklist

### DIFF
--- a/.github/workflows/ci-web.yml
+++ b/.github/workflows/ci-web.yml
@@ -50,7 +50,12 @@ jobs:
       - run: npm ci
       - run: npm run lint
       - run: npm run format:check
-      - run: npm audit --audit-level=high
+      # Scoped to production dependencies: dev-only transitive advisories
+      # (eslint/vitest/workbox tooling) must not block application PRs. An
+      # unscoped gate deadlocked the repo for a month in 2026-08 — it failed
+      # every web PR, including the Dependabot PRs carrying the fixes, and no
+      # single bump could clear it. Anything reaching users is still gated.
+      - run: npm audit --omit=dev --audit-level=high
 
   test:
     name: Test & Coverage
--- a/.github/workflows/e2e-web.yml
+++ b/.github/workflows/e2e-web.yml
@@ -8,8 +8,10 @@ name: E2E — Web Smoke
 #
 # The job runs in the official Playwright container (browsers + OS deps baked
 # in) instead of `npx playwright install --with-deps` every run. The image tag
-# MUST match the @playwright/test version in web/package-lock.json — bump them
-# together (a version mismatch means untested browser builds).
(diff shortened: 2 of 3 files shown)


skizzy1986:
Internal change — no release note. This PR only touches CI config (scoping the `npm audit` gate to production deps and deriving the Playwright container tag from the lockfile) and bumps transitive lockfile entries; there's no user-visible behavior change.

---

PR #173, opened 2026-08-18
fix: live PM5 erg session save writes real schema columns and never loses a session

### LINKED ISSUE #174: fix: live PM5 erg session save writes nonexistent columns — no bluetooth session has ever landed (Sprint 5 · T1)
**Sprint 5 — Benchmark Integrity, T1 (Size M, Week 1, no dependencies).** Full spec context: [`coach/SPRINT_BENCHMARK_INTEGRITY.md`](https://github.com/skizzy1986/erg-dashboard/blob/claude/benchmark-improvement-sprint-6b51ll/coach/SPRINT_BENCHMARK_INTEGRITY.md)

**Story:** As Scott, I want a session I complete on the live PM5 view to actually save to my training log, so that a rowed benchmark (or any live session) isn't silently lost.

**The bug (DB-verified 2026-07-16, re-verified on main @ 514520d 2026-08-18):** `ErgLiveView.jsx:490-491` inserts `watts`/`distance` and an ISO `YYYY-MM-DD` date, but `sessions` has only `avg_watts`/`avg_hr`/`distance_m` with text `MM/DD/YY` dates. Production has **zero** rows with `source='bluetooth'` or `status='logged'` — the path has never worked. The offline queue (`useOfflineQueue.js`) re-inserts the same broken shape.

**Acceptance criteria**
1. Online save inserts with real schema columns (`avg_watts`, `avg_hr`, `distance_m`) and the row appears in the session list within one query refresh.
2. `date` is written as `MM/DD/YY` text, matching every other row.
3. Queued offline rows drain with the same corrected shape.
4. Non-offline insert failures (RLS, missing `user_id`) queue the row and show "saved locally, will sync" rather than claiming success.
5. Null/undefined PM5 summary fields are omitted/null-safed without throwing.
6. A regression test asserts the exact row shape passed to `supabase.from('sessions').insert(...)`.

**Edge cases:** null PM5 fields; RLS vs network failure; drain idempotency; pre-fix malformed rows already in the local queue.

**Out of scope:** Concept2 auto-import (#56), ErgLiveView UI redesign, bluetooth backfill (none exists).

skizzy1986:
Closes #174

## What changed and why

The live PM5 save path inserted columns that don't exist on `sessions` (`watts`/`distance`) with an ISO date, so no bluetooth session ever landed in production and failures were silently swallowed by an offline queue that retried the same broken shape forever — this fixes the row shape, makes the queue self-heal, rescues manually-ended sessions, and makes every save outcome visible (Sprint 5 · T1 · #182).

Details:
- Row now writes `avg_watts`/`distance_m`, unpadded `M/D/YY` local-time date via a new `toLogDate()` in `utils/dateFormat.js`, `duration` as `MM:SS`; `avg_hr` omitted (the PM5 General Status packet carries no HR)
- Fixed the `todaySession` matcher (ISO-vs-`M/D/YY` compare + `normType` mangling made it always null) and suffixed labels with end time so same-day sessions don't collide on `UNIQUE(date, label)`
- Offline queue normalizes legacy rows on read (storage self-heals on first drain), treats duplicate-key `23505` as delivered (lost-ack recovery), and guards against concurrent drains
- Save flow has observable states (`saving`/`saved`/`queued`/`failed`) with reset gated behind DONE; enqueue wrapped in a quota guard; pending-sync count on the connect screen
- Manual END builds a summary via `usePM5.finish()` instead of discarding the session
- Also carries the Sprint 5 plan doc (`coach/SPRINT_BENCHMARK_INTEGRITY.md`, approved at Gate 1)

Follow-up filed: #184 (pre-existing read-side date bugs that become visible once this list is non-empty).

## How to test manually

Covered by CI (38 new tests, 428 total). For a live check: connect a PM5 in the Erg Live view, row a piece, end it manually with END, save — the session should appear in the log with watts/distance populated; airplane-mode saves should show "SAVED LOCALLY — WILL SYNC" and drain on reconnect.

## Checklist

### DIFF
--- a/coach/SPRINT_BENCHMARK_INTEGRITY.md
+++ b/coach/SPRINT_BENCHMARK_INTEGRITY.md
@@ -1,0 +1,390 @@
+# Sprint Plan: Benchmark Integrity — 2026-08-18 → 2026-09-18 (re-timed)
+
+> **STATUS: GATE 1 — PROPOSED, AWAITING SCOTT'S APPROVAL.**
+> Originally produced 2026-07-16; re-timed 2026-08-18 after the original
+> window elapsed with the gate unanswered. All eight task findings were
+> re-verified against main @ `514520d` (2026-08-18) — every one still holds.
+> Nothing below is committed work: no spec, no code, no GitHub issues exist
+> yet. On approval, each task becomes an issue and runs through the full gated
+> pipeline (`/orchestrate`) individually.
+
+## What changed between 2026-07-16 and 2026-08-18
+
+- **The benchmarks slipped entirely.** The CP retest (session 61, 7/5) was
+  **cancelled**, never rescheduled. No 5k benchmark session exists in the DB.
+  The `anchors` table is untouched since 2026-07-01: `rowing_cp` still 205 W
+  provisional, `bike_ftp` still unvalidated, `current_block` still names the
+  July 8–21 block. A month of benchmark cadence was lost silently — exactly
+  the failure mode T2 below is designed to catch.
+- **The repo moved.** The App.jsx milestone landed (`erg-dashboard.jsx`
+  dissolved); #146/#147/#151/#154 merged (doc sweep, coverage ratchet to
+  68/67/64, `useSessionLog` extraction, THEME tokens); the `npm audit` CI
+  blocker was resolved and the dependabot backlog cleared (2 PRs remain, both
+  green with auto-merge armed). None of this touched the benchmark surface —
+  re-verified by grep on 2026-08-18 (see task list).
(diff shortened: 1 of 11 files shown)


skizzy1986:
**Release note (Fixed)**

Live PM5 erg sessions now actually save. Previously, rowing sessions logged from the Erg Live view could silently fail to appear in your training log — this is fixed, along with a same-day session naming collision and a save-status indicator (Saving / Saved / Queued / Failed) so you always know where a session stands. Sessions saved while offline (e.g. airplane mode) now sync reliably once you're back online, and manually ending a session with END now saves it instead of discarding it.

---

PR #186, opened 2026-08-20
fix: erg session list read-side date handling, plus /daily brief and doc accuracy pass

### LINKED ISSUE #184: fix: erg session list read-side date bugs — text ordering + ISO-only date_display (D-5 from #174 spec)
Pre-existing read-side defects found during the #174 spec (Sprint 5 / #182), filed per Gate 2 decision. These become **user-visible for the first time** once #174 ships, because the `status='logged'` erg list has been empty until now.

1. `useErgSessions.js:39` orders by `date` as **text** — with `M/D/YY` values, `"9/1/26"` sorts above `"10/1/26"`. Ordering needs to go through a parsed date (client-side parse or `to_date(date,'MM/DD/YY')`-equivalent).
2. `enrich` (`useErgSessions.js:15`) splits the date on `-` (ISO-only), so `date_display` silently falls back to the raw string for every `M/D/YY` row — i.e. for all real data.

Scope: read path only (`useErgSessions.js` + tests). Reuse `utils/dateFormat.js` helpers (`toISODate`, and `toLogDate` once #174 lands) — no new date formatters per the one-helper rule.

Not blocking #174's merge, but should land soon after — the wrong ordering shows up the moment the list is non-empty.

skizzy1986:
Closes #184

## What changed and why

Fixes the erg read path's ISO-date assumption against the live `M/D/YY` TEXT column (#184), adds a `/daily` skill that generates a 30-minute task list from live state, and corrects five stale claims in `CLAUDE.md` / `README.md`.

Three commits, independently reviewable:

**1. `feat(skills): add /daily`** — a read-only skill that reads open PRs and their CI status, the issue backlog, the `anchors` table and recent vitals, then returns 3–6 tasks that fit ~30 minutes, led by a single "if you only do one thing" pick. It writes nothing to the repo or DB. Gating rule: while `current_phase` is `maintenance`, the brief never surfaces training adherence, missed-session counts, CTL/ATL/TSB decline or streaks.

**2. `fix: erg session list read-side date handling (#184)`** — `sessions.date` is TEXT `M/D/YY` and the erg read path assumed ISO throughout:

- **Ordering** — the server-side `.order('date')` sorts the TEXT column lexically, so `"12/31/25"` ranks above `"1/1/26"`. The enriched list is re-sorted on the normalized ISO key via the existing `toISODate()` helper, matching the pattern already shipped in `views/mobile/MobileAnalytics.jsx`.
- **`date_display`** — `split('-')` only matched ISO, so on live `"6/13/26"` input both parts came back `undefined` and the chart axis silently fell back to the raw string.
- **`withinLast7Days`** in `ErgView` built a `Date` from the raw string, which is Invalid Date for `M/D/YY` — the SESSIONS / WK counter was permanently `0`.

Not fixed, deliberately: the `.limit(50)` is still applied to the lexical order server-side, so past 50 rows the wrong 50 are fetched. Noted in a comment at the call site. There are 0 rows matching this query today (`status='logged'` is written only by the PM5 save path, live as of #173), so the limit is far from binding; fixing it properly needs a server-side ordering expression.

Coverage floor ratcheted **68/67/64 → 72/71/68** (measured 76.86 / 75.45 / 72.00).

**3. `docs: correct stale claims`** — each verified against the repo rather than the prose: React 18 → 19 (4 occurrences, `package.json` pins `^19.2.8`); App.jsx is 447 lines not ~515 (2 occurrences); the coverage-exclude paragraph read as though the former monolith were excluded, which `vite.config.js` explicitly contradicts, and omitted `test-setup.js`; the structure block omitted `src/services/`, `src/components/mobile/`, `src/views/program/`; Edge Functions listed 1 of the 4 that exist. `ProgramView.jsx` and `StrengthLogger.jsx` line counts were checked and left alone — both already accurate.

## How to test manually

Covered by CI. The six new tests were each confirmed to **fail** against the unfixed source and pass with it, so they exercise the bug rather than the fix.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_015yFLjAqk4QqRTDNt96kQyH)_

### DIFF
--- a/.claude/skills/daily/SKILL.md
+++ b/.claude/skills/daily/SKILL.md
@@ -1,0 +1,72 @@
+---
+name: daily
+description: Generate a short list of tasks that genuinely fit a 30-minute window, ranked from live repo and health state. Use when Scott has a gap and wants to know what to pick up without reading the backlog.
+argument-hint: (none)
+---
+
+Produce a list of 3–6 tasks that each land inside ~30 minutes, ranked from live
+state. Read-only — gather, rank, print. Write nothing to the repo or the DB.
+
+> Focus arrives in short, unpredictable windows. The cost of deciding what to
+> work on can exceed the window itself, so this skill makes the decision.
+
+## Usage
+
+```
+/daily
+```
+
+## Step 1 — Gather live state (run in parallel, all read-only)
+
+| Source | Query | Feeds |
+|---|---|---|
+| GitHub | `list_pull_requests` open, then `pull_request_read` `get_check_runs` on each | Green + mergeable PRs — always rank first |
+| GitHub | `list_issues` open, P1 first | Backlog candidates |
+| Supabase | `select key, value from anchors where superseded_at is null` | **Gates the health section — read this before writing anything** |
+| Supabase | `select * from vitals order by date desc limit 7` | Which health fields are going unrecorded |
+| Bash | `git branch --show-current`, `git status --porcelain` | Uncommitted work, wrong-branch warnings |
+| Bash | `ls web/node_modules \| head -1` | Whether deps are installed |
+
+Project id for Supabase is in `CLAUDE.md`. Never guess a task's scope — if the
+code hasn't been checked, spawn an `Explore` agent to size it, or label it
+"needs a look first" and give no estimate.
+
+## Step 2 — Rank
+
+Highest value-per-minute first:
+
+1. **Finished work awaiting one action** — a green, mergeable PR. The work is
+   already done; only the click is missing. Always the top candidate.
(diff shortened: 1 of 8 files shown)


skizzy1986:
**Release notes draft**

**Fixed:** Erg session history now sorts correctly — sessions used to be ordered incorrectly around year boundaries (e.g. late December showing after early January) because of how dates were stored. Session dates on the Erg tab now display properly instead of appearing blank, and the "Sessions / Wk" counter now correctly counts recent sessions instead of always reading 0.

*(The `/daily` skill and documentation corrections in this PR are internal/dev-tooling changes with no end-user-facing effect, so they're omitted above.)*

---

Issue #77, opened 2026-06-29
refactor: decompose Program tab into views/program/* (≈3 PRs, next month)


skizzy1986:
**Overflow from the month plan — the single largest seam (~3,071 lines), too big for one PR.** Split across ~3 PRs, each via `/refactor`:
1. `views/ProgramView.jsx` shell + `views/program/ProgramPhases.jsx` (PHASES, ANNUAL_ARC).
2. `views/program/ProgramMicrocycle.jsx` (2-week roster).
3. `views/program/ProgramYear.jsx` (EVENT_PATHWAY year view).

Each lands a render test and removes its files from the C1 coverage `exclude`. Per the architecture decision recorded in H0 (#67). Part of #52.

PR #190, opened 2026-08-21
fix: make the benchmark badge actually clear when the benchmark is done

skizzy1986:
Closes #175

## What changed and why

The benchmark due/overdue indicator shipped in #189 with a deterministic defect: completing a benchmark did not clear its badge — the exact silent failure #175 exists to prevent, inverted. This fixes that, plus two further defects found while validating the fix.

**The defect.** `findMatch` returned the first eligible session in payload order, and the payload is ordered by the TEXT `M/D/YY` column, so `'7/5/26'` sorts above `'6/23/26'`. Combined with `searchEnd = max(window.end, today)`, CP Test #1's search range ran ~50 days past its own one-day window and consumed the retest belonging to CP Test #2:

```
before:  [0] quiet   claimed=61   <- CP Test #1 steals the retest
         [1] overdue claimed=null <- CP Test #2 never clears
after:   [0] quiet   claimed=45
         [1] quiet   claimed=61
```

Match selection is now deterministic and independent of payload order: prefer a session inside the entry's own window, then earliest date, tie-break on lowest id. Nearest-to-window was considered and rejected — session 61 (`7/5`) is *nearer* to the `7/1` window than session 45 (`6/23`), so the steal would have survived it. Choosing by date rather than position also removes this feature's exposure to the ordering bug in #187.

**A 5k logged the way it is actually logged now clears.** The keyword was `5k`; the real labels are `5,000m` and `PM — 5,000m`. A comma between digits no longer splits a token, and keywords expand k-notation to metre-notation, so `5,000m`, `PM — 5,000m`, `5000m Time Trial` and `5k TT` all clear.

**A ride's elevation no longer clears a rowing benchmark.** That comma rule also made a Zwift label's trailing elevation field match — `… 1:12:30, 22.4km, 1,000m` tokenised to `1000m` and satisfied the 1000m benchmark, and 2,000m of climbing satisfied the 2k. A rowed distance is written `5,000m` and never carries a `km` token; every ride row does, so a `km` token now disqualifies a candidate.

**`Row-2k` keeps its `2k` token.** Dropping the letter-hyphen-digit fold restores conformance with the approved spec, which only ever specified the digit-then-letter direction. `4-min` → `4min` still folds; `130-142W` still splits.

## Test changes, which are the point of this PR

Every test that could have caught the defect resolved a **one-entry ladder**, which removes the entry that does the stealing, and the view test mocked the hook outright. All 19 of those tests pass against the buggy code.

[...]

PR #189, opened 2026-08-20
Add benchmark status resolver and calendar badges

skizzy1986:
Closes #175

## What changed and why

Scheduled benchmark tests were slipping silently. The CP retest was cancelled on 7/5 and never rescheduled (session 61), the end-of-base 5k never happened at all, and `anchors` has been untouched since 7/1. The event ladder rendered as static text, so neither gap surfaced anywhere in the app.

This adds an `OVERDUE` / `DUE` badge to the ladder panel in `CalendarView`. On today's data it produces exactly two badges — CP Test #2 and the 5k Time Trial — which are precisely the two tests that slipped.

**New files**

| File | Layer | Purpose |
| --- | --- | --- |
| `utils/eventWindow.js` | utils | Parses fuzzy ladder dates (`~Mid Jul 26`, `12 Sep–9 Oct 26`) into calendar windows; keyword derivation; timezone-safe day arithmetic |
| `utils/benchmarkStatus.js` | utils | Pure resolver: ladder + sessions → per-entry status, with grace window and claim-once |
| `hooks/useBenchmarkSessions.js` | hooks | Scoped query, key `['sessions','benchmark-window']` |
| `hooks/useBenchmarkStatuses.js` | hooks | Composes query + resolver; the only clock read in the feature |
| `components/BenchmarkBadge.jsx` | components | Presentational pill, inline styles, no data access |

`views/CalendarView.jsx` gains two imports, one hook call, and one JSX element (+14 lines).

## Design decisions worth reviewing

`EVENT_LADDER` entries carry **no link to a `sessions` row** — no id, no label key, no date join. So "was this benchmark done?" is answered heuristically, and the details matter:

- **Whole-token keyword matching**, not substring. Under substring matching a `24min` recovery row would satisfy the `4min` CP test.
- **14-day backward grace.** Not cosmetic — session 61, the only row representing the CP retest, is dated six days *before* its own window opens, so a strict in-window search could never clear it. It also prevents a false overdue on CP Test #1, genuinely completed eight days early.
- **Claim-once.** CP tests #1 and #2 share the `cp` keyword with overlapping grace ranges; without this, one session clears both and masks a genuinely missed test.
- **`cancelled` and `planned` read as NOT done.** `COMPLETED_STATUSES` is imported, never redefined. The `loggedSessions` prop is deliberately unused here — `useSessionLog`'s `status !== 'planned'` filter puts cancelled rows into it, which would invert the whole feature.
[...]

PR #191, opened 2026-08-21
feat: benchmark due/overdue indicator (Sprint 5 · T2)

### LINKED ISSUE #175: feat: benchmark due/overdue indicator — stop tests slipping silently (Sprint 5 · T2)
**Sprint 5 — Benchmark Integrity, T2 (Size M, Week 1, no dependencies).** Full spec context: [`coach/SPRINT_BENCHMARK_INTEGRITY.md`](https://github.com/skizzy1986/erg-dashboard/blob/claude/benchmark-improvement-sprint-6b51ll/coach/SPRINT_BENCHMARK_INTEGRITY.md)

**Story:** As Scott, I want the app to tell me when a scheduled benchmark test (CP retest, 5k, future 2k) is due or overdue, so that a test doesn't silently slip the way the CP retest did in July (session 61 cancelled 7/5, never rescheduled; the end-of-base 5k never happened at all — anchors untouched since 7/1).

**Acceptance criteria**
1. An `EVENT_LADDER` entry with `kind: 'benchmark'` whose date has passed and whose linked session is `planned`/`cancelled` (or absent) shows a visible "overdue" signal (surface chosen at Spec stage).
2. A benchmark within the next 7 days shows a distinct "upcoming" signal.
3. Concrete cases that must work on first load today: the CP revalidation and the end-of-base 5k both read as overdue.
4. Completing the linked session clears the signal via normal query invalidation.
5. Free-text dates (`'~Early Aug 26'`) degrade gracefully to "upcoming, exact date TBD".

**Edge cases:** unparseable dates; no linked session; `cancelled` must read as NOT done; `MM/DD/YY` text-date comparison; year-on-year repeats not double-flagged.

**Out of scope:** writes to `anchors`; restructuring `EVENT_LADDER`/`ANNUAL_ARC`; the ProgramView split (#77/#114) — prefer a small standalone component over editing `ProgramYear.jsx` mid-split.

skizzy1986:
Closes #175

## What changed and why

Surfaces overdue and upcoming benchmark tests so one cannot silently slip the way the CP retest did in July — session 61 was cancelled on 7/5 and never rescheduled, the end-of-base 5k was due in early August and never logged, and `anchors.rowing_cp` has sat at 205 provisional since 7/1 with nothing on screen saying so.

On first load today this surfaces three overdue benchmarks: CP Test #1, CP Test #2, and the 5k Time Trial.

### How it works

Four new modules, composed into a self-contained banner:

| File | Role |
|---|---|
| `utils/eventLadderDates.js` | Resolves an `EVENT_LADDER` entry to a date window |
| `utils/benchmarkStatus.js` | Derives overdue / upcoming / done / pending / none |
| `hooks/useBenchmarkSessions.js` | React Query read of completed sessions |
| `components/BenchmarkStatusBanner.jsx` | Renders the signals |

`EVENT_LADDER` gains three optional fields on its five `kind: 'benchmark'` entries — `windowStart`, `windowEnd`, `matchTerms`. Every pre-existing value on all nine entries is unchanged, and the four non-benchmark entries are untouched.

`CalendarView.jsx` gains exactly two lines, an import and `<BenchmarkStatusBanner />`. The banner calls its own hook rather than taking sessions as a prop, so nothing threads through and `ProgramYear.jsx` is not touched at all — the #77/#114 split is clear of this change.

### Design decisions worth flagging for review

**A benchmark is only "done" if a session matching `COMPLETED_STATUSES` exists.** A `cancelled` session reads as not-done. Session 61 is the regression fixture: it label-matches and falls inside the date tolerance, so the status gate is the only thing rejecting it, and a test proves that by flipping just its status and watching it match.

**Match tolerance is asymmetric — 3 days before the window, 7 days after** — encoding that benchmarks slip late rather than happening early. This is why session 45 (`6/23/26`, a real 4-min CP test eight days before the ladder slot) does not clear CP Test #1. Deliberate: the 1-minute companion piece was cancelled, so CP still cannot be recomputed.

**Absence of evidence is never completion.** A missing, unreadable, or RLS-empty session list leaves a benchmark still-due. There is no code path where "no session found" becomes done.

**The component has no affirmative all-clear state.** It renders overdue and upcoming, or nothing. This is structural rather than cosmetic — while loading it cannot flash a false cleared state, because there is no cleared state to flash.

### Follow-up commit — review findings

The initial matcher used a raw substring test, which matched `5k` inside `18.5km` and `1000m` inside `21000m`. A false match renders nothing and is permanent, so it recreates precisely the silent-slip failure this feature exists to prevent. Matching is now boundary-guarded, and `\b` is deliberately not used — in `18.5km` the `.` is itself a word boundary, so `\b` leaves the bug intact. No regex lookbehind appears anywhere in the module, since iOS Safari below 16.4 lacks support.

Thousands separators are now stripped before matching. This closed a live gap: `5,000m` contains neither `5k` nor `5000m`, so a benchmark logged in the existing house style (sessions 19, 20, 25) would never have cleared the indicator.

A failed session query now renders a neutral `BENCHMARK STATUS UNAVAILABLE` line rather than nothing, so an outage cannot be mistaken for "nothing due". The loading state still renders nothing, so a slow first fetch cannot flash that line before the real signals arrive.

`1-min` was dropped from the tune-ups match terms — a one-minute piece appears in warm-ups, CP tests and intervals, and is not evidence the benchmark happened.

Verified against all 92 live session rows: of 460 label/term combinations, the only 7 hits are genuine benchmark labels.

## How to test manually

Open the Calendar tab. Three overdue benchmarks should appear above the `UPCOMING EVENTS · SEASON 1 LADDER` heading. No competition, TARGET, or optional ladder entry should ever appear there.

## Checklist

509 tests pass (up from 484). New files sit at 100% lines and functions; branch coverage 96.66% on `benchmarkStatus.js` and 100% elsewhere. Three per-file 80/80/70 gates added; no threshold lowered.

## Deferred to follow-up issues

- **Reschedule/acknowledge state.** The three overdue benchmarks will render indefinitely, since they genuinely never happened. Recommended design is a `planned` session row rather than a dismiss button, so a reschedule is real and visible rather than hidden in one device's local storage.
- **`useSessions.js`** orders a TEXT `MM/DD/YY` column lexically under `.limit(50)`, so `'12/1/26'` sorts before `'7/1/26'` and the returned rows are effectively arbitrary. Pre-existing; avoided here via a dedicated hook rather than fixed.
- **`useSessionLog.js:69`** treats `status !== 'planned'` as done, classifying cancelled sessions as complete. Pre-existing; deliberately not copied.
- **`useOfflineQueue.js:73`** inserts sessions and invalidates no query at all.

---
_Generated by [Claude Code](https://claude.ai/code/session_017iDtU2Bm6JTKwxWyijfn4K)_

### DIFF
--- a/web/src/components/BenchmarkStatusBanner.jsx
+++ b/web/src/components/BenchmarkStatusBanner.jsx
@@ -22,15 +22,58 @@ function shortDate(date) {
   return `${date.getDate()} ${MONTH_LABELS[date.getMonth()]}`;
 }
 
-// Two visible states only: OVERDUE and UPCOMING. There is deliberately no
+function containerStyle(compact) {
+  return {
+    background: THEME.surfaceAlt,
+    border: `1px solid ${THEME.divider}`,
+    borderRadius: 6,
+    padding: compact ? '8px 10px' : '12px 14px',
+    marginTop: 14,
+  };
+}
+
+function Heading() {
+  return (
+    <div
+      style={{
+        fontSize: 9,
+        letterSpacing: 2,
+        color: THEME.gold,
+        marginBottom: 8,
+      }}
+    >
+      BENCHMARKS
+    </div>
+  );
+}
+
+// Two signal states only: OVERDUE and UPCOMING. There is deliberately no
 // affirmative "all clear" state — with nothing to say the component renders
 // null, so an in-flight or failed session query cannot flash a false "done".
+// The one non-signal state, BENCHMARK STATUS UNAVAILABLE, is a negative
+// statement about the DATA and asserts nothing about the benchmarks.
 export default function BenchmarkStatusBanner({
   entries = EVENT_LADDER,
   today,
   compact = false,
 }) {
   const { data, isLoading, isError } = useBenchmarkSessions();
+
+  // Evaluated before derivation, so exactly one code path reaches the
+  // unavailable state and it can never co-render with a signal row. React
+  // Query v5 sets isError only when the query holds no successful data, so a
+  // background refetch failure cannot blank an already-correct banner.
+  if (isError) {
+    return (
+      <div style={containerStyle(compact)}>
+        {!compact && <Heading />}
+        <div style={{ fontSize: 9, letterSpacing: 2, color: THEME.muted }}>
+          BENCHMARK STATUS UNAVAILABLE
+        </div>
+      </div>
+    );
+  }
+
   const rows = deriveBenchmarkStatuses(entries, data ?? [], {
     today,
     sessionsReady: !isLoading && !isError,
@@ -42,27 +85,8 @@ export default function BenchmarkStatusBanner({
   if (visible.length === 0) return null;
 
   return (
-    <div
-      style={{
-        background: THEME.surfaceAlt,
-        border: `1px solid ${THEME.divider}`,
-        borderRadius: 6,
-        padding: compact ? '8px 10px' : '12px 14px',
-        marginTop: 14,
-      }}
-    >
-      {!compact && (
-        <div
-          style={{
-            fontSize: 9,
-            letterSpacing: 2,
-            color: THEME.gold,
-            marginBottom: 8,
-          }}
-        >
-          BENCHMARKS
-        </div>
-      )}
+    <div style={containerStyle(compact)}>
+      {!compact && <Heading />}
       {visible.map((r, i) => {
         const overdue = r.status === 'overdue';
         const accent = overdue ? THEME.orange : THEME.gold;
--- a/web/src/components/BenchmarkStatusBanner.jsx
+++ b/web/src/components/BenchmarkStatusBanner.jsx
@@ -1,0 +1,120 @@
+import { EVENT_LADDER } from '../constants/schedule.js';
+import { THEME } from '../constants/theme.js';
+import { useBenchmarkSessions } from '../hooks/useBenchmarkSessions.js';
+import { deriveBenchmarkStatuses } from '../utils/benchmarkStatus.js';
+
+const MONTH_LABELS = [
+  'Jan',
+  'Feb',
+  'Mar',
+  'Apr',
+  'May',
+  'Jun',
+  'Jul',
+  'Aug',
+  'Sep',
+  'Oct',
+  'Nov',
+  'Dec',
+];
+
+function shortDate(date) {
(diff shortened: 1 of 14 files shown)


skizzy1986:
Closing as a duplicate. #175 was already delivered by #189 and #190 and closed on 2026-08-20; this branch rebuilt the same feature without checking the issue tracker first. `coach/SPRINT_BENCHMARK_INTEGRITY.md` still reads "no GitHub issues exist yet", which is stale and is what this was scoped from.

Beyond the duplication, the implementation on `main` is better on the merits.

**`main` handles a case this branch gets wrong.** Zwift rows end `<time>, <dist>km, <elevation>m`. Stripping thousands separators — which this branch does so `5,000m` matches the `5000m` term — also turns an elevation of `1,000m` into `1000m`, which then clears the 1000m tune-ups benchmark. Verified against this branch's own matcher:

```
tune-ups: FALSE MATCH   "Zwift Alpe du Zwift — 1:12:00, 22.4km, 1,000m"
tune-ups: FALSE MATCH   "Zwift climb portal — 48:00, 15.2km, 1000m"
2k Test:  FALSE MATCH   "Zwift route — 40:00, 2,000m elevation"
```

`main` avoids this with `hasRideDistance()`: a label carrying a `km` token is a ride, so its trailing metres field is elevation rather than a rowing distance. That is exactly the false-DONE class this feature exists to prevent, and this branch reintroduces it.

`main` also has **claim-once assignment** (#16942fc), so a session clearing one benchmark cannot clear another. This branch lets a single CP retest row clear both CP Test #1 and #2, which masks a genuinely missed test.

The two also diverge on match tolerance, and `main`'s reasoning is sounder. This branch used −3/+7 on the theory that benchmarks slip late; `main` uses a 14-day backward grace because session 61 is dated *before* its own window opens, so a strict in-window search could never clear it.

**Worth keeping from this branch:** the outage state. `BenchmarkBadge.jsx` renders nothing when the session query fails, so a Supabase outage is indistinguishable from "no benchmarks due" — the feature's own failure mode arriving by a different door. Following up with a small PR against current `main` adding a neutral `BENCHMARK STATUS UNAVAILABLE` line.

Also filed from this work: #192 (reschedule/acknowledge state), #193 (`useSessions` lexical sort under `limit(50)`), #194 (`useSessionLog` counts cancelled as completed — 32 live rows), #195 (offline queue drain invalidates nothing).

---

Issue #195, opened 2026-08-21
Offline queue drain invalidates no queries


skizzy1986:
Found while building #175. Pre-existing, not introduced there.

## Problem

`web/src/hooks/useOfflineQueue.js:73` inserts sessions when the queue drains on reconnect, and calls `invalidateQueries` for nothing at all — not `['sessions']`, not `['tss-history']`, not `['benchmark-sessions']`.

So a session saved offline and flushed on reconnect does not refresh any view until a query's `staleTime` lapses (60s where set) or the app remounts. The row is in the database; the UI just doesn't know.

This matters most for the live PM5 path, since that is the one most likely to be used away from connectivity — and a benchmark rowed offline is exactly the case where the user most wants confirmation it landed.

## Fix

Mirror the pattern already used at `LogSessionForm.jsx:86` and `ErgLiveView.jsx:649`, invalidating on successful drain:

```js
queryClient.invalidateQueries({ queryKey: ['sessions'] });
queryClient.invalidateQueries({ queryKey: ['benchmark-sessions'] });
```

Consider whether `['tss-history']` and `['erg-sessions']` belong too — worth an audit of which query keys a session insert should invalidate, since the three existing write paths currently disagree with each other.

Related: `['tss-history']` appears never to be invalidated from anywhere, which looks like the same class of staleness bug.

## Note

#175 added `['benchmark-sessions']` invalidation to the two in-app write paths, but this third path was outside its scope.

skizzy1986:
**Triage**

- **Priority: Medium** — no data is lost (the row lands in the DB), but the UI gives no confirmation until a `staleTime` lapse or remount, which is worst precisely on the live PM5/offline path this issue calls out — the case where a user most needs to trust that a benchmark rowed offline actually landed. Straightforward fix (mirror the existing `invalidateQueries` pattern from `LogSessionForm.jsx`/`ErgLiveView.jsx`), so cheap to close relative to its trust impact.
- **Duplicate check:** none found. Related to but distinct from #175 (which added `['benchmark-sessions']` invalidation to the two in-app write paths but explicitly left this offline-drain path out of scope).

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #194, opened 2026-08-21
useSessionLog treats cancelled sessions as completed


skizzy1986:
Found while building #175. Pre-existing, not introduced there.

## Problem

`web/src/hooks/useSessionLog.js:69`:

```js
const loggedSessions = allSessions.filter((e) => e.status !== 'planned');
```

"Not planned" is treated as "done". But `cancelled` is neither — and there are **32 cancelled rows** in the table right now, against 58 completed. That is a third of the log potentially counted as training that did not happen.

The proof case is session 61: `CP RETEST — 1min + 4min max (rested, fed)`, `7/5/26`, `cancelled`. Under this filter it reads as a completed CP retest. It is the exact session whose cancellation went unnoticed for a month and prompted #175.

## Correct primitive

`web/src/constants/sessionStatus.js` already exports the right thing:

```js
export const COMPLETED_STATUSES = ['actual', 'completed', 'logged'];
```

`useTSSHistory.js` and `MobileAnalytics.jsx` already use it. `useSessionLog` is the outlier.

## Scope

`useSessionLog` feeds `App.jsx:187` → `CalendarView` and others, so this changes what renders. Worth checking whether anything downstream has quietly compensated for the current behaviour before flipping it.

Possible knock-on: cancelled sessions may currently be inflating training-load or session-count displays. Worth a before/after comparison rather than a blind swap.

## Note

#175 deliberately avoided copying this pattern — `matchBenchmarkSession` gates on `COMPLETED_STATUSES` with a comment naming the anti-pattern, and a regression test proves session 61 does not clear a benchmark.

skizzy1986:
**Triage**

- **Priority: High** — `useSessionLog`'s `status !== 'planned'` filter currently counts all 32 `cancelled` rows as completed, against 58 genuinely completed rows. That's a third of the visible log potentially misrepresenting training that didn't happen, feeding `CalendarView` and (per the issue) possibly training-load/session-count displays. This is the same class of bug that caused the month-long silent-slip which prompted #175, so it's actively producing wrong output today, not just a design gap.
- **Duplicate check:** none found. `matchBenchmarkSession` (added in #175) already gates on `COMPLETED_STATUSES` correctly and is not affected — this issue is specifically about `useSessionLog` being the remaining outlier.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #193, opened 2026-08-21
useSessions() orders a text date column lexically under limit(50)


skizzy1986:
Found while building #175. Pre-existing, not introduced there.

## Problem

`web/src/hooks/useSessions.js`:

```js
.order('date', { ascending: false })
.limit(50)
```

`sessions.date` is TEXT `MM/DD/YY`, so this is a **lexical** sort, not chronological. `'12/1/26'` sorts before `'7/1/26'` because `'1' < '7'`.

Combined with `.limit(50)` on a 92-row-and-growing table, the returned rows are an arbitrary and unpredictable subset — not "the 50 most recent". Any consumer assuming recency is silently wrong, and the failure is invisible: no error, just missing rows.

This is why #175 introduced a dedicated `useBenchmarkSessions` hook rather than reusing this one — a benchmark match falling outside the returned window would have produced a false OVERDUE.

## Fix options

1. **Order by `created_at`** (a real timestamptz) where insertion order is an acceptable proxy.
2. **Order in JS** after fetching, via `to_date`-equivalent parsing — but the `.limit()` still truncates arbitrarily server-side, so the limit must go first.
3. **Server-side expression order** — `.order()` cannot take `to_date(date,'MM/DD/YY')` through PostgREST directly; would need a view or a generated column.
4. **Migrate `sessions.date` to a real `date` type.** The correct long-term fix and the only one that removes the whole class of bug — but it touches every read and write path, so it needs its own story.

Option 4 is the real answer; options 1–3 are containment.

## Check before fixing

Audit what currently consumes `useSessions()` and whether any of it assumes chronological ordering — the fix may change on-screen behaviour, so it wants tests characterising current output first.

skizzy1986:
Duplicate of #187, which covers the same bug and does it better — it has the measured impact (21 of the 63 sessions on or after 2026-06-17 missing from the payload, and session 45 sitting at rank 68 where it never reaches the client at all), the full consumer list, and the cross-month failure mode I didn't cover.

Closing this one. #187 is the issue to work.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #192, opened 2026-08-21
Benchmark reschedule/acknowledge state (T2 follow-up)


skizzy1986:
Follow-up to #175 / #191, deferred at the review gate.

## Problem

CP Test #1, CP Test #2 and the 5k are overdue because they genuinely never happened, so they render indefinitely — roughly 197 days by the time the Jan 2027 windows open. A banner showing the same three warnings for six months becomes wallpaper, and alarm fatigue is how the original silent-slip bug recurs.

## Recommended design

Reschedule via a **`planned` session row**, not a dismiss button.

The capability already half-exists: a `planned` row can be put on the calendar today. The only missing piece is that the banner does not look at it. So the change is a *read*, not a write.

An overdue benchmark with a future `planned` label-matching session renders `↻ RESCHEDULED` with the new date, replacing the `⚠ OVERDUE` badge while still stating `Nd overdue` (ruled at the review gate).

Why this shape:

- **Nothing ever goes silent.** The row still renders, still names the benchmark, still carries `daysOverdue`. The invariant that no benchmark disappears is preserved.
- No dismissal concept is invented — nothing to sync, nothing device-local, nothing hidden from Coach.
- The remedy becomes "do the thing" rather than "hide the thing".
- Zero schema change, zero migration.

Rejected alternatives: `localStorage` (per-device, and invisible to Coach, who reads `sessions` via MCP and would keep re-prescribing — it recreates invisible state in a new place); `coach_log` (an append-only narrative with `supersedes` chains, wrong shape for queryable UI state, and Code authoring rows in Coach's content lane); a dedicated table (needs a migration and RLS policy for a cosmetic problem).

## Implementation sketch

- `constants/sessionStatus.js` — export `PLANNED_STATUS`.
- `hooks/useBenchmarkSessions.js` — widen to `.in('status', [...COMPLETED_STATUSES, PLANNED_STATUS])`.
- `utils/benchmarkStatus.js` — new `matchPlannedSession(entry, sessions, today)`, gated on `status === PLANNED_STATUS`, requiring `sessionDate >= today` (a past plan is a missed plan and must not suppress the warning — live sessions 43 and 56 are exactly that case), with **no** proximity window, since a reschedule is outside the original window by definition.

**Containment requirement:** `matchBenchmarkSession`'s status gate must stay `COMPLETED_STATUSES.includes(...)` so a `planned` row can never produce `done`. Two functions, two gates, never merged.

[...]

PR #196, opened 2026-08-21
fix: say when benchmark status is unavailable instead of rendering silence

skizzy1986:
Follow-up to #189 / #190.

## What changed and why

A failed sessions read resolves every ladder entry to `unknown` (`benchmarkStatus.js:91`), and `BenchmarkBadge` returns `null` for unknown. So an outage silently removes the overdue warnings and leaves the ladder panel looking reassuring — silence is exactly what "no benchmarks due" looks like.

That is the feature's own failure mode arriving by another door: the whole point is that a benchmark cannot slip unnoticed, and a Supabase hiccup currently makes three overdue benchmarks disappear.

The panel now carries a neutral line when the read has failed:

```
BENCHMARK STATUS UNAVAILABLE
```

It is a statement about the *data*, not about the benchmarks. No `⚠`, no `●` — reusing either would make an outage read as a signal — muted grey rather than the red/gold the badges own, and no claim that anything is fine.

**Pending is deliberately not unavailable.** A slow first read must not flash the outage line and then replace it with badges, so the line is driven by `isError` alone and the loading path still renders nothing.

## Why the hook lives where it does

`useBenchmarkDataUnavailable` sits alongside `useBenchmarkStatuses` rather than being read from `useBenchmarkSessions` in the view. Importing `useBenchmarkSessions` into `CalendarView` directly pulls `supabaseClient` into every test that renders the view, which breaks `CalendarView.test.jsx` at import time with `supabaseUrl is required` — it mocks the statuses module, not the sessions one. Keeping the view on a single hooks module preserves that boundary. It shares the query key, so react-query serves it from cache rather than issuing a second request.

## How to test manually

Open the Calendar tab with the network blocked (DevTools → Offline, or block the Supabase host). The ladder still lists its five entries, with `BENCHMARK STATUS UNAVAILABLE` under the heading and no badges. Restore the network and the line disappears as the badges resolve.

## Checklist

549 tests pass across 50 files (up from 544/50).

Three tests added to `CalendarView.benchmark.test.jsx`, which mocks only at the Supabase boundary so the real hook, resolver and badge all run:

- a failed read renders the line, keeps the ladder, and shows no badge
- a successful read shows no line
- the existing pending test now also asserts the line is absent

**The outage test has teeth** — verified by reverting the render condition to `false` and confirming it goes red, then restoring it.

## Related

Found while reviewing the benchmark work. Also filed: #192 (reschedule/acknowledge state), #193 (`useSessions` lexical sort under `limit(50)`), #194 (`useSessionLog` counts cancelled as completed — 32 live rows), #195 (offline queue drain invalidates nothing).

---
_Generated by [Claude Code](https://claude.ai/code/session_017iDtU2Bm6JTKwxWyijfn4K)_

### DIFF
--- a/web/src/hooks/useBenchmarkStatuses.js
+++ b/web/src/hooks/useBenchmarkStatuses.js
@@ -16,3 +16,14 @@ export function useBenchmarkStatuses(ladder = EVENT_LADDER, options = {}) {
     [ladder, data, today, sessionsReady]
   );
 }
+
+// A failed read resolves every entry to `unknown`, which renders no badge —
+// and silence is exactly what "no benchmarks due" looks like. Callers use this
+// to say the status is unknown rather than imply everything is fine. Pending is
+// deliberately NOT unavailable: a slow first read must not flash an outage line
+// and then replace it with badges. Same query key as above, so react-query
+// serves it from cache rather than issuing a second request.
+export function useBenchmarkDataUnavailable() {
+  const { isError } = useBenchmarkSessions();
+  return isError;
+}
--- a/web/src/views/CalendarView.jsx
+++ b/web/src/views/CalendarView.jsx
@@ -1,6 +1,9 @@
 import WorkoutItem from '../components/WorkoutItem.jsx';
 import BenchmarkBadge from '../components/BenchmarkBadge.jsx';
-import { useBenchmarkStatuses } from '../hooks/useBenchmarkStatuses.js';
+import {
+  useBenchmarkStatuses,
+  useBenchmarkDataUnavailable,
+} from '../hooks/useBenchmarkStatuses.js';
 import {
   getRosterMode,
   resolveDay,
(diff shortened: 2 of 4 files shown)


skizzy1986:
**Release note — Fixed**

The upcoming-events ladder on the Calendar tab no longer goes silent when benchmark status can't be loaded. If a data outage or network hiccup prevents your overdue/due benchmarks from being checked, you'll now see a "BENCHMARK STATUS UNAVAILABLE" note instead of the panel just looking clear. This prevents a real outage from being mistaken for "nothing's due."

---

skizzy1986:
**Release note draft**

**Fixed** — The Calendar tab's benchmark ladder no longer goes silently reassuring when the data can't load. If a benchmark status read fails, you'll now see "BENCHMARK STATUS UNAVAILABLE" under the ladder instead of the overdue/due badges just disappearing — so a connection hiccup can't be mistaken for "nothing's due." Loading still shows nothing while the first read is in flight, and the line clears automatically once the data comes back.

---

PR #198, opened 2026-08-21
fix: order sessions by a real date column so LIMIT stops dropping recent rows (#187)

### LINKED ISSUE #187: fix: sessions.date text ordering + limit(50) silently drops recent sessions from useSessions()
Found during the Stage 3 spec for #175 (Sprint 5 · T2), verified against live data on 2026-08-20. Filed as its own issue because fixing it was out of scope for #175 — that PR only works around it locally.

## The bug

`web/src/hooks/useSessions.js` does `.order('date')` on `sessions.date`, which is a **TEXT** column holding `M/D/YY`. Lexical ordering on that text is wrong in two distinct ways:

- **Within a month** — `'6/4/26' > '6/30/26'` lexically, so single-digit days sort above the end of their own month.
- **Across months** — `'9/30/26' > '12/1/26'`, so once the calendar crosses into October–December 2026, December sessions sort below September ones.

Combined with `.limit(50)`, this does not merely return rows in a strange order — it returns **the wrong set of rows**.

## Measured impact (92-row table, 2026-08-20)

**21 of the 63 sessions on or after 2026-06-17 are missing from the payload.**

Actual ranks under the current text-desc order:

```
rank 40,41,42 → 7/1/26      rank 43,44,45 → 6/9/26      rank 50 → 6/5/26
rank 52,53,54 → 6/4/26      rank 55       → 6/30/26   ← excluded by limit 50
```

`6/30/26` falls outside the 50 while `6/4/26` sits inside it.

Most consequentially, session 45 (`CP Test - 4min MAX (GATED)`, `6/23/26`, `completed`) sits at **rank 68** and never reaches the client at all.

## Affected consumers

`useSessions()` has three live consumers, all currently reading an arbitrary subset of recent history:

- `web/src/hooks/useCoach.js:81`
- `web/src/views/mobile/MobileSessionLog.jsx:34`
- `web/src/views/mobile/MobileAnalytics.jsx:34`

This degrades as rows accumulate (~1.2/day), and gets structurally worse when the calendar crosses into Q4 2026.

## Suggested fix

Order by the parsed date rather than the text. Per the data-layer convention in `CLAUDE.md`, chronological ordering of `sessions` is `to_date(date,'MM/DD/YY')` — which means either a database-side ordering (an RPC or a generated/indexed date column) or client-side sorting through `toISODate` from `web/src/utils/dateFormat.js` after fetching a sufficient window. Reuse the existing helper; per the one-helper rule, no new date formatter.

Worth deciding at spec time whether the durable fix is a generated `date_iso` column on `sessions` (additive, reversible, indexable) rather than repeated client-side normalisation — that would also close the read-side half of #184.

## Relationship to other issues

- **#175** works around this with a separate scoped query (`['sessions','benchmark-window']`, `.limit(500)`) so its own correctness does not depend on the broken ordering. That workaround is deliberate and local; it does not fix this.
- **#184** covers the same class of read-side date bug in `useErgSessions.js` (text ordering + ISO-only `date_display`). These two should probably be specced together.

skizzy1986:
Closes #187

## The bug

`sessions.date` is a **TEXT** column holding unpadded `M/D/YY`, so ordering on it is lexical — wrong within a month (`'6/4/26' > '6/30/26'`) and across months (`'9/30/26' > '12/1/26'`). Combined with `.limit(50)` in `useSessions()`, that returns the **wrong set of rows**, not merely a strange order.

Measured against the live 92-row table:

| | |
|---|---|
| Genuinely-recent sessions the old query dropped | **8** |
| Stale sessions it included instead | **8** |

## The fix

A generated `date_iso` date column derived from the text `date`, indexed on `(user_id, date_iso desc nulls last)`, with `useSessions` and `useTSSHistory` ordering on it. The text `date` column is untouched — every write path still writes `M/D/YY` and every display still reads `sessions.date`, so no consumer changed.

### Three things worth reviewing carefully

**`to_date` could not be used.** Both the issue and `CLAUDE.md` suggested `to_date(date,'MM/DD/YY')`. It is `provolatile = 's'` (STABLE — it honours `DateStyle`), so Postgres rejects it in a generated column *and* in an index expression; `text::date` is STABLE too. The parser uses `make_date`/`split_part`, which are genuinely IMMUTABLE, and agrees with `to_date` on **92/92 live rows**.

**The parser returns NULL rather than raising.** It is recomputed on every `UPDATE`, so a raising expression would abort a Coach-written session outright and leave any bad row permanently un-updatable. A `NOT VALID` CHECK makes bad *writes* loud instead — so `M/D/YY` is now enforced by the schema rather than by convention in a markdown file, including for Coach's MCP writes. `NOT VALID` means existing rows are never re-checked, which keeps the rollback unconditional.

**`nullsFirst: false` is mandatory, not stylistic.** `postgrest-js` emits the nulls clause *only* when the option is defined (`dist/index.cjs:881`). A bare `.order('date_iso', {ascending: false})` sends `date_iso.desc`, which is Postgres default **NULLS FIRST** — an unparseable date would sit at the top of the window and present as the most recent session. That would be worse than the bug being fixed.

## Scope notes

**`useTSSHistory` is included for consistency, not correctness.** I initially believed its misordering corrupted CTL/ATL/TSB. It does not: `calcTrainingLoad()` keys a map by `toISODate(date)`, derives min/max with `reduce` rather than by array position, and walks day-by-day. Input order cannot affect the output — **CTL/ATL/TSB have been correct all along.** What was genuinely wrong is its ordering test, which fed an already-sorted fixture through a mock that ignores `.order()`; that has been replaced with a real contract assertion.

**`useErgSessions.js` and `useBenchmarkSessions.js` are deliberately untouched.** Both already self-correct client-side, and their comments remain accurate while they keep ordering on `date`. Moving them onto `date_iso` is a one-line follow-up each.

**The dead `useSessions()` call in `useCoach.js` is removed.** The value was never read or returned, but it mounted a live query observer that refetched on every `['sessions']` invalidation. Separately worth knowing: `buildTrainingContext()` is never called outside its own test file, so the Coach training context is unreachable in production — a follow-up, not something to wire up inside a bug fix.

## Verification

Migration applied via `apply_migration`, then verified against live data:

```
parser:  6/4/26→2026-06-04  6/30/26→2026-06-30  9/30/26→2026-09-30
         12/1/26→2026-12-01  12/31/25→2025-12-31  2026-08-07→2026-08-07
         2/31/26→NULL (no error)  garbage→NULL
backfill: 92 total / 92 parsed / 2026-05-22 → 2026-08-07 / 0 stale
proof:    recent rows the old query dropped = 8
          stale rows it included            = 8
```

Supabase advisors: the one finding this change introduced (`function_search_path_mutable`) is fixed — the function now pins `search_path = ''`. All remaining advisor findings are pre-existing and unrelated.

Gates: **558 tests pass** across 51 files, lint clean, build exits 0. Coverage rose from 76.86/75.45/72.00 to **79.92/77.47/74.49**, so the global floor is ratcheted 72/71/68 → **75/73/70**, and `useSessions.js` (previously untested, 100% lines now) plus `useTSSHistory.js` get per-file 80/80/70 entries.

## Rollback

`supabase/migrations/008_sessions_date_iso_rollback.sql`. **Revert the deploy first** — a client carrying `.order('date_iso')` against a schema without the column gets a PostgREST 400 on every sessions read.

### DIFF
--- a/.claude/skills/erg-context.md
+++ b/.claude/skills/erg-context.md
@@ -33,7 +33,9 @@ Testing Library. Full briefing: `CLAUDE.md`.
 ## Supabase gotchas (honour on every write)
 
 - Supply `user_id` explicitly on inserts. RLS is always on — never bypass it.
-- `sessions.date` is TEXT `MM/DD/YY` — order with `to_date(date,'MM/DD/YY')`.
+- `sessions.date` is TEXT `M/D/YY` — **write** it, never order by it. Order by the
+  generated `date_iso` date column (`nullsFirst: false` on desc). `date_iso` is
+  read-only; an insert naming it is rejected.
 - `UNIQUE(date, label)` on `sessions` — temp-suffix labels before bulk shuffles.
 - Vitals upsert on `(user_id, date)`. Migrations are additive and reversible;
   DDL via `apply_migration`, never raw `execute_sql`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -122,7 +122,11 @@ coach use), and the context store (Code↔Coach shared memory, added by #94):
 | `coach_log`         | **Context store** — append-only diary + decision record (Coach's content) |
 | `anchors`           | **Context store** — current calibration + phase state (one live row/key)  |
 
-**`sessions` columns:** `date` (**text `MM/DD/YY`**), `type`, `label`, `duration`,
+**`sessions` columns:** `date` (**text `M/D/YY`, unpadded** — the write column),
+`date_iso` (**generated `date`, read-only** — derived from `date` by
+`session_date_to_iso()`, indexed, and the only correct thing to `order by`; an
+insert that names it is rejected, and a `date` that will not parse is rejected by
+the `sessions_date_parseable` check), `type`, `label`, `duration`,
 `srpe`, `prs`, `exercises` (jsonb), `coach_note`, `status`, `coach_flag`,
 `avg_watts`, `avg_hr`, `distance_m`, `source` (default `portal`; Coach writes
 `coach`), `user_id`. **No watt-target columns — targets live in `label` +
@@ -171,7 +175,10 @@ legacy `sessions.date` text pattern); RLS single-owner policy like the modern ta
 **Data-layer gotchas (honour on every write):**
 - Supply the `user_id` UUID explicitly on inserts — `auth.uid()` is the column
   default but does not resolve through the MCP connector.
-- Order `sessions` chronologically with `to_date(date,'MM/DD/YY')` (date is text).
+- Order `sessions` chronologically with **`date_iso`** (generated + indexed), never the
+  text `date` — lexical order silently returns the wrong rows under a `LIMIT`. On
+  descending, pass `NULLS LAST` (`nullsFirst: false` from supabase-js). Do not use
+  `to_date(...)`: it is STABLE, so it cannot be indexed or used in a generated column.
 - `UNIQUE(date, label)` on `sessions` — temp-suffix labels before bulk shuffles
   (`set label = label || '~tmp'`).
 - Vitals upsert: `on conflict (user_id, date) do update`.
--- a/supabase/migrations/008_sessions_date_iso.sql
+++ b/supabase/migrations/008_sessions_date_iso.sql
@@ -1,0 +1,69 @@
+-- 008_sessions_date_iso.sql
+-- Additive + reversible. Gives sessions a real DATE key derived from the legacy
+-- TEXT `date`, so PostgREST .order() is chronological and .limit() selects the
+-- right rows (#187). The TEXT column is untouched: every write path still writes
+-- "M/D/YY" via toLogDate(), every display still reads sessions.date.
+--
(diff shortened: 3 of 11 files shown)


skizzy1986:
**Draft release note — Fixed**

Your training calendar and dashboard could occasionally show older sessions in place of your most recent ones. This happened because sessions were sorted using their text-formatted date rather than a true calendar date, which silently mis-ordered rows whenever a view was limited to a fixed number of entries (recent workouts could be pushed out of view while stale ones took their place). Sessions are now ordered by a proper date field, so recent training always shows up first. Note: CTL/ATL/TSB fitness numbers were never affected by this bug — they were calculated correctly all along.


---

skizzy1986:
**Release note (Fixed)**

Fixed: the dashboard was sometimes showing the wrong set of recent sessions — up to 8 genuinely recent sessions could be missing while older ones appeared in their place. This happened because sessions were being sorted by date as text instead of as a real date, which sorts incorrectly across month and year boundaries (e.g. `6/30` could rank below `6/4`, and `12/1` below `9/30`). Session lists now sort correctly, so the most recent sessions always show. Training load (CTL/ATL/TSB) was not affected by this bug.


---

PR #201, opened 2026-08-21
config: sync SplitIQ's shared UI layer to claude.ai/design

skizzy1986:
## What

Adds the [design-sync](https://claude.ai/design) inputs so Claude Design builds mockups from SplitIQ's **real** components instead of generic ones — and so future re-syncs are reproducible rather than hand-redone.

Design project: https://claude.ai/design/p/8edb6124-729f-4d6c-9e32-6ba955e2a6b4

## Why it needed this much wiring

This repo is an **app, not a component library** — no library build, no `dist/` entry, no `.d.ts`, no Storybook. The converter's defaults assume a published package, so the gaps are closed by config rather than by editing any app source:

| File | Closes |
|---|---|
| `entry.jsx` | Components are default exports; the converter's synth entry uses `export *`, which drops defaults. This barrel gives them named exports. |
| `gen-css.mjs` → `base.css` | The app ships no stylesheet (inline styles + runtime-injected vars). This generates the token layer + app ground from `theme.js`. |
| `config.json` `dtsPropsFor` | Plain JSX yields no extractable types — ts-morph emitted `[key: string]: unknown` for all 11. Prop contracts are hand-written from the real destructured params. |
| `docs/*.md` | Per-component usage docs; `category:` frontmatter sets the DS pane groups. |
| `previews/*.tsx` | 33 authored preview cells built from the components' own tests and the real training model. |
| `conventions.md` | Prepended to the generated README and inlined into the design agent's prompt: the no-classes inline-style idiom, the token vocabulary, and what each accent *means*. |

## Verification

- `package-validate.mjs` exits 0; render check **11/11 clean** — 0 bad, 0 thin, 0 floor cards.
- Every preview cell graded against the absolute rubric; all `good`.
- Re-run of the driver reports **11/11 grades carried forward, 0 cleared** — the next sync will be incremental.

## Scope

Touches **no application code**. Everything is under `web/.design-sync/` plus five `.gitignore` lines for generated output. `LogSessionForm` is deliberately excluded — it constructs a Supabase client at module scope, which throws in a browser IIFE bundle; the reason and the fix are recorded in `NOTES.md`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,10 @@ playwright-report
 .playwright
 # Lighthouse CI local run output
 .lighthouseci/
+
+# design-sync (claude.ai/design) — staged scripts, build output, machine state
+web/.ds-sync/
+web/ds-bundle/
+web/.design-sync/.cache/
+web/.design-sync/learnings/
+web/.design-sync/node_modules
--- a/web/.design-sync/DESIGN_BRIEF.md
+++ b/web/.design-sync/DESIGN_BRIEF.md
@@ -1,0 +1,657 @@
+---
+title: UI & Information Architecture Brief
+category: guidelines
+status: proposed
+date: 2026-08-22
+---
+
+# SplitIQ — UI & Information Architecture Brief
+
+Companion to [`conventions.md`](./conventions.md). That document is **descriptive** — written to make
+a design agent reproduce the current look, not to state intent. This one is **prescriptive**: where
+SplitIQ should go, and why.
+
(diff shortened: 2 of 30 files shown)


skizzy1986:
Internal change — no release note. This PR only adds Claude Design sync config/tooling under `web/.design-sync/` (component docs, previews, prop typing, generated token CSS) plus a few `.gitignore` lines; it touches no application code and has no user-visible effect.

---

PR #202, opened 2026-08-21
feat: compute training load from power when sRPE was never captured

skizzy1986:
Your CTL/ATL/TSB have been running on almost nothing since **3 July**.

`useTSSHistory` filtered on `.gt('srpe', 0)`, and `srpe` is null on **42 of 58** completed sessions — every Strava-backfilled row, because sRPE is subjective and cannot be recovered after the fact. Those rows were dropped in SQL before they ever reached the client, so only **2 of the 16** sessions since 1 July fed the training-load curve.

## The fix

A new `sessionLoad()` in `web/src/utils/trainingLoad.js`: use sRPE when it exists, otherwise fall back to power — intensity factor (`avg_watts / threshold`) squared per the standard TrainingPeaks model, scaled so an hour at threshold reads 10.

That scaling is the important part. The existing formula produces "sRPE-hours" (an hour at sRPE 7 → 7), so the power value has to land on the *same* unit or the CTL curve becomes a mix of two incompatible scales. An hour at threshold reading 10 is the sRPE-10 equivalent.

Rowing measures against **rowing CP (205 W)**, bike against **bike FTP (250 W)** — both already live in `anchors`. Using CP for both would overstate every ride by ~50%.

The SQL filter becomes `.or('srpe.gt.0,avg_watts.gt.0')`, which is the actual reason the rows were missing.

## Effect on live data, 1 July onward

| Date | Type | Before | After |
|---|---|---:|---:|
| 7/01 | erg (sRPE 7) | 7 | **7** |
| 7/01 | cycling 196 W | 0 | 8 |
| 7/03 | erg (sRPE 7) | 5 | **5** |
| 7/11 | erg 142 W | 0 | 4 |
| 7/13 | erg 159 W | 0 | 5 |
| 7/14 | erg 147 W | 0 | 5 |
| 7/16 | erg 166 W | 0 | 7 |
| 7/20 | strength | 0 | 0 |
| 7/21 | cycling (Zwift, no power) | 0 | 0 |
| 7/24 | erg 89 W | 0 | 1 |
| 7/26 | erg 155 W | 0 | 4 |
| 7/27 | erg 118 W recovery | 0 | 1 |
| 7/30 | erg 143 W | 0 | 4 |
| 8/02 | erg 153 W | 0 | 4 |
| 8/04 | erg 151 W | 0 | 4 |
| 8/06 | erg 152 W | 0 | 5 |

14 of 16 restored. **The two rows that already carried sRPE are unchanged** — existing history is preserved exactly, not recomputed. The remaining zeros are the strength session and the Zwift ride, neither of which has power data at all.

## One judgement call worth challenging

On the only two sessions carrying *both* signals (150 W, sRPE 7), the power model reads **~24% lower** — 5.35 against 7. I did **not** calibrate that away.

At 73% of CP an RPE of 7 is genuinely elevated, and this project's own coaching note on the 7/3 session reads exactly that gap as fatigue ("endurance by power despite RPE7; RPE + HR drift read as fatigue"). Fitting a scaling factor to close it on n=2 would convert a fatigue signal into a calibration constant and destroy it permanently. The consequence is a small step in CTL where hand-rated sessions meet power-derived ones — visible, but honest.

If you'd rather they line up, that's a one-line change to the multiplier and worth a deliberate decision.

## Notes

- Load is derived in a `useMemo` rather than the `queryFn`, so revalidating CP recomputes the whole series without a refetch (same pattern as `useErgSessions`).
- A missing anchor degrades to 0 rather than falling back to a hardcoded threshold — matching `useAnchors`' existing "never silently reuse a stale constant" contract. Covered by a test.
- **574 tests pass**, lint clean, build exits 0. Coverage: `trainingLoad.js` 100% lines, `useTSSHistory.js` 100% lines. Global lines 79.92 → **80.14**.

### DIFF
--- a/web/src/hooks/__tests__/useTSSHistory.test.js
+++ b/web/src/hooks/__tests__/useTSSHistory.test.js
@@ -6,6 +6,7 @@ import React from 'react';
 const fromMock = vi.fn();
 const inMock = vi.fn();
 const orderMock = vi.fn();
+const orMock = vi.fn();
 
 vi.mock('../../supabaseClient.js', () => ({
   supabase: {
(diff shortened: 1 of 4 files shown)


skizzy1986:
**Fixed — Training load (CTL/ATL/TSB) was undercounting most of your recent sessions.**

Any completed session without a manually-entered sRPE — including every Strava-imported ride and erg row — was being silently dropped from the fitness/fatigue calculation. Those sessions with recorded power now contribute to training load too, estimated from power relative to your rowing CP or bike FTP so it lines up with the existing sRPE-based scale. Sessions you've already rated by sRPE are untouched, and sessions with neither sRPE nor power (e.g. strength, no-power rides) still read as zero.


---

skizzy1986:
**Release note (Fixed)**

Your training load chart (CTL/ATL/fitness trend) was silently skipping most rowing and cycling sessions imported from Strava — 42 of the last 58 completed sessions never fed the curve, because they don't carry a subjective effort (sRPE) rating and the calculation required one. It now falls back to your session's power output when sRPE isn't available, so the fitness/fatigue trend reflects the training you actually did, not just the sessions you hand-rated. Sessions that already have an sRPE score are unchanged.

---

PR #203, opened 2026-08-21
fix: reconcile displayed CP with the live anchor (#176)

skizzy1986:
Closes #176 (Sprint 5 · T3).

## Problem

`anchors.rowing_cp` has been 205 W provisional since 07-01, but the app kept showing 190 W — a seed constant — as the current CP. `ErgView` already read the anchor correctly for the CP tile and pace bands; the contradiction came entirely from static copy sitting alongside it.

## What changed

| Surface | Before | After |
|---|---|---|
| `CRITICAL_POWER.cpEstimate` | `190` | removed — CP is only ever read live |
| `PACE_ZONES` (static export) | bands derived from the 190 seed | removed — a second set of bands that could diverge from the live path |
| `CALIBRATION_STATUS` load-model row | `'CP/FTP est. 190W (untested), …'` | resolved at render from the anchor |
| `FTP_TEST.when` | `'… (~Jun 24+)'` | undated; states the test is overdue |
| `App.jsx` | 4 unused imports of these constants | removed |

New pure resolver in `constants/trainingConfig.js`:

```js
deriveCalibrationStatus({ cp, cpAvailable, cpStatus })
```

`ErgView` feeds it straight from `useAnchors()`. Available → `CP 205W (provisional), strength TSS ±30%, CTL needs 42d`. Unavailable → `CP anchor unavailable, …` — never a silent 190.

`PACE_ZONES`' only consumer outside the module was `PaceTrendChart.test.jsx`, which now derives its own fixture.

## Acceptance criteria

1. ✅ Conflicting static copy is live-anchor-driven or undated.
2. ✅ Unreachable anchor reads "unavailable" — no silent 190 fallback (`useAnchors` already returns `cp: null`; the new row text follows it).
3. ✅ No new hardcoded numeric CP.
4. ✅ No remaining `190` presented as a current CP.

## Deliberately out of scope

- **Confidence tiers stay put.** Upgrading `conf`/`tier` when a benchmark lands is #181, and there's a test asserting this PR does *not* do it.
- **`POWER_DURATION.predW` untouched** (incl. the 30-min row's `~180–190`). That table belongs to #180, and re-deriving the ranges off a new CP is a training-science call, not a display fix. The `190` there is a predicted test wattage, not a current-CP claim — the `ErgView` test comments explain why the assertions are scoped around it.
- No writes to `anchors`.

## Verification

- `npm test` — **595 passed** (52 files), up from 576; +19 new.
- `npm run lint` — 0 errors (3 pre-existing warnings).
[...]

PR #204, opened 2026-08-22
fix: build the Coach's training context on the client and actually send it (#199)

### LINKED ISSUE #199: fix: the Coach training context is never built — buildTrainingContext is unreachable in production
Found while removing a dead `useSessions()` call in #198 (fixing #187).

`buildTrainingContext()` is exported from `web/src/hooks/useCoach.js:8` and is thoroughly tested, but **it is never called outside its own test file**. `useCoach` does not invoke its own exported function, and its return object (`messages, streamingContent, isStreaming, error, model, setModel, sendMessage, clearHistory, vitals, tssQuery`) contains no context string.

So the Coach currently receives none of the training context the function is designed to assemble — load, vitals, readiness, recent sessions, today's plan.

`useCoach.js:81` used to call `useSessions()` and discard the result, which looks like staging for exactly this wiring. #198 removed that call rather than wiring it up, because deciding what the Coach should see is a product question with real surface — how many sessions, whether planned rows are included, what lands in the prompt — and it did not belong inside a bug fix.

Scope for this issue: call `buildTrainingContext` from `useCoach`, decide and document what feeds `recentSessions` (note `useSessions()` now returns a correct, chronologically-ordered 50 after #187), and cover the wiring with a test that asserts the context actually reaches the request.

skizzy1986:
Closes #199

## What changed and why

`buildTrainingContext` is now called by `useCoach` and posted to `coach-chat`, and the edge function's own ~90-line `buildContext` is deleted rather than repaired.

### The diagnosis in #199 was incomplete

The issue is right that the Coach gets no context, but the client-side dead code is only half of it. `coach-chat` has its **own** `buildContext` that runs on every request — and all four of its sections are broken against the real schema, so it returns the literal string `CURRENT TRAINING DATA (as of 2026-08-22):` and nothing else. Verified against production (`swdrueaserjzhuxnzmeu`, deployed v9):

| Section | Why it never renders |
|---|---|
| `TSB / CTL / ATL` | Filters `.eq('status','logged')` — **zero** of the 92 session rows have that status (35 `actual`, 23 `completed`, 32 `cancelled`, 2 `planned`). Even with rows it would read 0: `tssMap` is keyed by the TEXT `M/D/YY` date while the accumulator loop looks up ISO keys. |
| `Readiness / RHR / HRV / Sleep` | Selects `date, rhr, hrv, sleep` from `vitals`. Those columns **do not exist** — they are `rhr_bpm, hrv_ms, sleep_hours`. PostgREST 400s, and `const [{ data: vitalsRows }] = await Promise.all(...)` destructures only `data` and never checks `error`, so the failure is silent. |
| `Today's session` | Compares `s.date === today` — TEXT `"8/22/26"` against ISO `"2026-08-22"`. |
| `Recent sessions` | Same absent `'logged'` status, over a top-10 ordered by the lexically-sorting TEXT `date` (the #187 bug). |

### Why the client owns it now

Every one of those calculations already exists on the client, correct and under test: `sessionLoad`'s power fallback (#202), `calcTrainingLoad`, `COMPLETED_STATUSES`, `computeReadiness`, the live CP/FTP anchors, and `useSessions`' `date_iso` ordering (#187). Repairing the Deno copy would have been a third implementation of the same maths in a file with no tests and outside the coverage gate — which is exactly how it rotted unnoticed. `coach-chat` is now a thin Anthropic proxy holding no data logic.

### Bugs fixed in the builder itself

Three defects survived only because the existing tests fed synthetic data:

- **`sessions.duration` is free text** (`"45:00"`, `"60min"`, `"1h4m"`), so `` `${s.duration}min` `` rendered `45:00min`. Now parsed via the existing `parseDurationMinutes`.
- **Recent-session dates printed raw `M/D/YY`** next to an ISO header, leaving day/month order ambiguous to the model. Both are ISO now.
- **The header date came from `toISOString()`**, i.e. UTC — which names *yesterday* for the whole Perth morning, precisely when readiness is checked and today's session asked about. Reads local calendar fields now, matching `toLogDate`'s documented rationale.

### Decisions taken (with Scott, before any code)

- **`recentSessions`**: `COMPLETED_STATUSES` only, capped at 8 — a training week's worth of rows. The filter matters most for `cancelled`, which would otherwise dominate the list at 32 rows.
- **`todayPlanned`**: matched with `toISODate` on both sides.
- **Scope**: wiring only. Driving the prompt off live state is #209.

## How to test manually

Open the Coach tab and ask "How's my training load looking?" — the reply should quote real CTL/ATL/TSB and name recent sessions rather than giving generic advice. Cross-check the numbers against the Analytics tab. In devtools, the `coach-chat` request body should carry a populated `context`.

**Deploy note:** the edge function is a separate manual step and is *not* done by merging. Ordering does not matter — an old cached PWA bundle against the new function sends no context and gets the base prompt; the new bundle against the old function has its context ignored. Both degrade to today's behaviour, neither errors.

Worth noting: the deployed v9 was pushed as a bare single file and has never matched the repo copy (it used the service-role key + an explicit `user_id` filter). That drift disappears with `buildContext`, and this deploy makes the repo the source of truth for the first time. There is no CI check for edge-function drift — filed as #210.

## Correction to an earlier version of this description

An earlier revision said the prompt's `Current phase: Base (Jun–Aug 2026)` "expires in 9 days". That was wrong. Checking the live `anchors` table shows it was **already superseded on 2026-08-20** — `current_phase` is now `maintenance`, confirmed, with the note *"Health and consistency only — no progression targets, no adherence measurement."* The drift is live, not pending, and it is broader than the CP number: phase, block, intensity rules, TID model and progression cadence in `COACH_SYSTEM_PROMPT` all contradict the anchors. That is now tracked as **#209**, which also gives the Coach the `coach_log` decision history so it knows *why* things changed.

This matters for sequencing: once this PR merges and the function deploys, the Coach starts receiving real data against a frozen doctrine — making it more confidently wrong, not less. #209 should follow closely.

## Checklist

- [x] CI is green (lint, test, build) — 587 tests pass, 0 lint errors, Prettier clean, build succeeds
- [ ] No unexpected console errors in the browser — needs the preview deploy

**Test coverage:** new `useCoachWiring.test.js` (8 cases) asserts the context actually reaches the `fetch` body — the issue's explicit requirement — plus status filtering, the 8-row cap, free-text duration parsing, and the `M/D/YY`-vs-ISO `todayPlanned` match as a direct regression test. It is a separate file from `useCoach.test.js` because that one mocks `@tanstack/react-query` wholesale and `vi.mock` is hoisted and file-scoped. Two `todayISO` cases pin the local-vs-UTC day under a faked Perth clock. `CoachView` gains its first badge test.

**Coverage gate:** `useCoach.js` measures 83.47 lines / 82.35 functions / 74.68 branches and takes the standard 80/80/70 per-file gate. Global floor ratchets 75/73/70 → 80/75/72.

**Accepted gap:** the edge function has no tests — there is no Deno harness in this repo and adding one is out of scope. After this change the only logic left in it is trim/truncate/concat; everything that can be wrong now lives in `useCoach.js` under the coverage gate.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -117,7 +117,7 @@ coach use), and the context store (Code↔Coach shared memory, added by #94):
 | `exercises`         | Exercise library — 873 rows, **text ids**                            |
 | `exercise_media`    | Demo media per exercise (reference content, not user data)           |
 | `exercise_prefs`    | Per-user, per-exercise preferences (e.g. rest seconds)               |
-| `coach_messages`    | Legacy/experimental in-app chat rail (see Coaching data model)       |
+| `coach_messages`    | In-app Coach chat rail (see Coaching data model)                     |
 | `backup_snapshots`  | Daily full-DB JSON snapshots (backup cron)                           |
 | `coach_log`         | **Context store** — append-only diary + decision record (Coach's content) |
 | `anchors`           | **Context store** — current calibration + phase state (one live row/key)  |
@@ -364,8 +364,18 @@ The **integration model was ratified 2026-07-01**: Coach (Claude in chat) operat
 natively via the **Supabase MCP connector, writing directly to the DB** (`source='coach'`)
 — reading vitals, and inserting/updating `sessions`, `strength_workouts`, etc. This
 is the live coaching rail, so **Code and Coach share one source of truth: this file
-plus the schema above.** The in-app Anthropic-API path (`coach_messages` table) was
-trialed and set aside — treat it as legacy/experimental unless revived.
+plus the schema above.**
+
+The in-app Anthropic-API path (`coach_messages` table, the Coach tab) was revived on
+2026-08-22 by #199. **The client assembles the training context and posts it**:
+`buildTrainingContext` in `web/src/hooks/useCoach.js` composes load, readiness,
+today's prescription and the recent microcycle from the same tested utilities the
+dashboard itself uses (`calcTrainingLoad`/`sessionLoad`, `computeReadiness`,
+`COMPLETED_STATUSES`, `toISODate`), and `coach-chat` is a thin Anthropic proxy that
+holds no data logic of its own. It used to rebuild that context server-side against
+column names `vitals` does not have and a `status` filter no row matched, so the
+Coach silently ran on an empty context — **do not reintroduce a second builder in the
+edge function.**
 
 **Bridge discipline persists** even though Code holds repo + schema + deploy: Scott
 authorises consequential/destructive/schema changes; review structure before material
--- a/supabase/functions/coach-chat/index.ts
+++ b/supabase/functions/coach-chat/index.ts
@@ -3,17 +3,15 @@
 // Required secrets (set via Supabase Dashboard → Edge Functions → Secrets):
 //   ANTHROPIC_API_KEY  — Anthropic API key
 //
-// SUPABASE_URL and SUPABASE_ANON_KEY are injected automatically by Supabase.
-//
 // Deployed with JWT verification ON (default). The caller must send
 // Authorization: Bearer <supabase-session-token>.
-
-import { createClient } from 'https://esm.sh/@supabase/supabase-js@2';
 
 const CORS_HEADERS = {
   'Access-Control-Allow-Origin': '*',
   'Access-Control-Allow-Headers': 'authorization, x-client-info, apikey, content-type',
 };
+
+const MAX_CONTEXT_CHARS = 8000;
 
 const MODEL_MAP: Record<string, string> = {
   haiku: 'claude-haiku-4-5-20251001',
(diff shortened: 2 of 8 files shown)


skizzy1986:
**Rebased onto `main` (`2ac5542`) after #203 and #208 landed — force-pushed, `bdd10d8` → `e0d447d`.**

The rebase was textually clean. Checking it by hand anyway turned up a semantic conflict it hid, so there is now a second commit with a real behaviour change.

### The conflict

#208 rescaled `sessionLoad` from sRPE-hours to classic TSS — an hour at sRPE 10 now reads **100**, not 10 — and recalibrated every downstream band to match. `App.jsx`, `MobileAnalytics` and `LoadTooltip` all cut at `>10 / >-10 / >-30`. Its own comment names `useCoach`'s GREEN/AMBER/RED among the thresholds calibrated for the new unit.

`buildTrainingContext`'s GREEN cut was still `> 5` — a pre-rescale figure. Left as-is, a TSB of +6 would have the Coach saying "GREEN" while every screen in the app showed the same number as neutral amber. One value, two answers, which is the exact class of drift this PR exists to remove.

Nothing failed. No file overlapped. Both branches were individually correct — the incompatibility only existed in the combination, which is why it needed reading rather than a green test run.

### Fix

GREEN boundary moved to `> 10` to match the dashboard, with both boundaries pinned by a test so the two cannot separate again. The `-30` band is a subdivision of "fatigued" that RED already covers, so the three-signal mapping is unchanged.

One existing test asserted GREEN at `tsb: 10`; under the corrected boundary that is AMBER, so it was updated deliberately rather than worked around.

### Numbers after the rebase

- **616 tests pass** (was 587 — #203/#208 brought 29 with them), 0 lint errors, Prettier clean, `npm run build` succeeds. Supersedes the 587 figure in the description checklist.
- `useCoach.js` still measures 83.47 lines / 82.35 functions / 74.68 branches, clearing its 80/80/70 per-file gate.
- Global floor still ratchets 75/73/70 → 80/75/72.

---

skizzy1986:
**Rebased again onto `main` (`c6d8ab0`) — force-pushed, `e0d447d` → `cd5947b`.** #212, #213 and #214 landed while this sat open.

This time the rebase touched a file I also changed, so it was checked rather than trusted.

### `vite.config.js` — overlapping edit, resolved as a union

#212 and #214 added four per-file coverage gates (`invalidateSessionQueries.js`, `useErgSessions.js`, `useBenchmarkSessions.js`, `CalendarView.jsx`) into the same thresholds block where this PR adds `src/hooks/useCoach.js`. Git replayed the hunks separately; verified afterwards that **both** sets survived and neither side clobbered the other. They are independent gates, not competing values.

The global floor is this PR's edit alone and is unchanged at 80/75/72.

### One improvement taken: `PLANNED_STATUS`

#214 added `PLANNED_STATUS` to `constants/sessionStatus.js`, with a comment framing `'planned'` as a first-class concept deliberately kept out of `COMPLETED_STATUSES`. `buildTrainingContext`'s today-lookup was still comparing against the bare literal, so it now uses the constant — that is what it exists for, and #209 stacks more planned-status logic on this same code.

No behaviour change; `COMPLETED_STATUSES` is untouched so the recent-sessions filter is unaffected.

### Numbers on `cd5947b`

- **661 tests pass** (was 616), 0 lint errors, Prettier clean, `npm run build` succeeds.
- `useCoach.js` measures 83.47 lines / 82.35 functions / 74.68 branches — clears its 80/80/70 gate.
- Global measured 84.75 / 79.55 / 77.16 against the 80/75/72 floor. Left as-is: the extra headroom came from the tests #212–#214 brought with them, and ratcheting the floor on another PR's work belongs to that PR, not this one.

### Note on the delay

That is two rebases in an afternoon, and the first one hid a real semantic conflict behind a clean replay — #208's TSS rescale left this PR's TSB bands on the old scale, which would have had the Coach and the dashboard disagreeing about the same number. Each further round of drift is another chance for something like that to slip through unnoticed, so this is worth merging sooner rather than later.

---

skizzy1986:
**Third rebase — onto `8ef673d` (#218). Force-pushed, `cd5947b` → `d50a9c2`.**

This one found a second semantic conflict, and a pointed one: **#218's own commentary names this code path.**

### What #218 changed

It closed the readiness fork and gave `computeReadiness` a new contract — `{ score, status, color, partial }` — where a missing RHR now returns `score: null`, `status: 'NO DATA'` instead of inventing a number. Its comment:

> Previously a missing row scored 0 FATIGUED and an all-null row scored 100 READY — a fabricated verdict in either direction, and **the 0 defeated CoachView's `!= null` guard so it reached the Coach prompt as fact.**

`useVitals` keeps `readinessScore` / `readinessLabel` as legacy aliases, so this PR compiled and every test passed. But `buildTrainingContext` interpolates the score directly into `` `${score}/100` ``, so after the rebase it would have written:

```
Readiness: null/100 NO DATA | RHR: — | HRV: 27ms | Sleep: 6.8h
```

That is the same class of bug #218 had just removed, reintroduced in a new place — and worse here, because this PR promotes readiness from a badge to the Coach's primary input. A model reading `null/100 NO DATA` is being handed a non-answer shaped like an answer.

### Fix

A null score now reports `Readiness: unavailable (NO DATA)` and still lists the raw RHR / HRV / sleep, so the model sees exactly what is known and no verdict that was never computed.

Also surfaces the new **`partial`** flag. #218 sets it when a score was computed without HRV or sleep and notes it is *"surfaced so a view can say so"* — the Coach is a consumer that should weight a partial score rather than read it as complete. It now renders as `Readiness: 72/100 CAUTION (partial — scored without HRV or sleep)`.

Three tests pin all three cases: null score, partial score, complete score.

### Numbers on `d50a9c2`

- **670 tests pass**, 0 lint errors, Prettier clean, build succeeds.
- `useCoach.js` 83.62 lines / 82.35 functions / 76.74 branches — clears its 80/80/70 gate.
- Global 84.46 / 79.30 / 76.90 against the 80/75/72 floor.

### Worth saying plainly

Three rebases this afternoon, and **two of them hid real semantic conflicts behind a clean textual replay** — #208's TSS rescale left the TSB bands on the old scale, and now #218's readiness contract. Neither produced a conflict marker, a failing test, or a red check. Both were only caught by reading what landed.

`main` is moving faster than this PR is being reviewed, and every round is another chance for one of these to go unnoticed. The branch is green, up to date and has no outstanding review comments.

---

skizzy1986:
**Release notes draft**

**Fixed:** The Coach chat now actually sees your real training data. Previously it silently fell back to generic advice — the context it needed (training load, readiness, today's session) never reached the model due to mismatched database columns and date formats behind the scenes. Ask the Coach about your training load, readiness, or today's session and it will now reference your real numbers (CTL/ATL/TSB, RHR/HRV/sleep, recent sessions) instead of giving generic advice.

*Note: this fix also requires a separate edge-function deploy step (tracked in the PR description) before it's live in production.*

---

Issue #207, opened 2026-08-22
S2 — Data truth: put the desktop dashboard on live data


skizzy1986:
Design-brief slice **S2** (`web/.design-sync/DESIGN_BRIEF.md` §1.3, §7). One of the two slices the brief says to do first.

## Why

The desktop dashboard's headline numbers are decorative:

| Location | What it does |
|---|---|
| `App.jsx:190` | CTL/ATL/TSB from the static `DAILY_TSS` constant, not `useTSSHistory` |
| `App.jsx:204` | `totalErgDist = 55000` — hardcoded |
| `App.jsx:166` | `ftp = 190` — hardcoded, and nothing ever calls `setFtp` |
| `App.jsx:85-122` | e1RM history for 8 lifts — hardcoded, passed to `StrengthView` |
| `trainingConfig.js:261` + `tssData.js:1` | Two copies of `DAILY_TSS`; desktop imported one, mobile the other |

Meanwhile `useTSSHistory`, `useVitals`, `useStrengthPRs` and `useAnchors` all exist, work, and are already wired on mobile.

The brief's argument (§1.3) is that this is a **design prerequisite**, not a separate chore: *"you cannot delete a number you do not trust."* Data-truth uncertainty is what drives the data hoarding behind the "too much on screen" complaint. S4 (the Today rebuild) must not build a prettier frame around fake numbers.

## Checklist

- [x] **S2a — training load.** Shipped in #208 (`2ac5542`).
- [x] **S2b — vitals.** Shipped in #218 (`8ef673d`).
- [ ] **S2c — strength PRs + CP display.** `StrengthView` onto `useStrengthPRs`; delete `ftp = 190`; resolve `totalErgDist`.

## Notes carried out of exploration

**The #202 follow-on (folded into S2a, shipped in #208).** PR #202 made power-only sessions count, but left a unit mismatch that only bites once desktop reads live data. `sessionLoad` returned "sRPE-hours" (~1–8/session) while every TSB band — `App.jsx:193`, `LoadTooltip`, `MobileAnalytics`, `useCoach`, `calcReadiness`'s `tsb < -20`, R4's `tsb < -25` — was calibrated for classic TSS (14–121).

[...]

Issue #176, opened 2026-08-18
fix: reconcile displayed CP 190 with live 205 anchor (incl. static PACE_ZONES) (Sprint 5 · T3)


skizzy1986:
**Sprint 5 — Benchmark Integrity, T3 (Size S, Week 2, no dependencies).** Full spec context: [`coach/SPRINT_BENCHMARK_INTEGRITY.md`](https://github.com/skizzy1986/erg-dashboard/blob/claude/benchmark-improvement-sprint-6b51ll/coach/SPRINT_BENCHMARK_INTEGRITY.md)

**Story:** As Scott, I want every screen that shows a CP/FTP number to match the current live anchor, so that I'm not making training decisions off a stale hardcoded estimate while a newer provisional number exists in the DB.

**Facts:** `trainingConfig.js:148` hardcodes `cpEstimate: 190`; `trainingConfig.js:326` exports a static `PACE_ZONES = derivePaceZones(CRITICAL_POWER.cpEstimate)` off that stale 190; live `anchors.rowing_cp` = 205 provisional. Per the ratified 2026-07-02 Coach decision, the app reads CP live from `anchors` — never a swapped-in constant.

**Acceptance criteria**
1. Conflicting static copy (`cpEstimate: 190`, static `PACE_ZONES` + consumers, the `CALIBRATION_STATUS` "190W (untested)" row, `FTP_TEST` past-dated copy) is made live-anchor-driven or explicitly marked historical.
2. Loading/unreachable anchor shows a neutral "unavailable" state — never a silent 190 fallback.
3. No new hardcoded numeric CP introduced.
4. Zero remaining "190"s presented as current CP in affected files.

**Edge cases:** React Query pending vs null must read differently; `confirmed` vs `provisional` display without new hardcoding; static-`PACE_ZONES` consumers must not diverge from the live `derivePaceZones(cp)` path in `ErgView.jsx:222-223`.

**Out of scope:** changing `useAnchors`; wiring CP into TSS/CTL/ATL; bike FTP unless trivially the same pattern.

PR #208, opened 2026-08-22
fix: put the desktop dashboard on live training load, in units its bands mean

skizzy1986:
First of three PRs for **S2 — Data truth** (#207). Also closes out the unresolved follow-on from #202.

## The problem

Desktop CTL/ATL/TSB came from a static `DAILY_TSS` array that stopped on **2026-06-13**. Every headline load number on the home screen was decorative, while mobile already read the live hook. Per the brief (§1.3), fixing this is a design prerequisite for S4, not a separate chore — *you cannot delete a number you do not trust.*

Two things were wrong, and fixing either alone would have shipped a bug.

### 1. The unit — the #202 follow-on

`sessionLoad` returned "sRPE-hours" (an hour at sRPE 7 → `7`), so live sessions score **1–8**. Every threshold downstream was calibrated for **classic TSS**, where the old static seed ran **14–121**:

| Site | Band |
|---|---|
| `App.jsx:193`, `MobileAnalytics`, `LoadTooltip` | `tsb > 10` / `> -10` / `> -30` |
| `useCoach.js:20` | `tsb > 5` GREEN, `> -10` AMBER |
| `analysis.js` | `tsb < -25` (R4), `tsb < -20` (readiness deduction) |

The bands were **~10× too wide**, and the effect is not subtle. Replaying all 31 signal-carrying sessions through both scales:

| | Old (sRPE-hours) | New (classic TSS) |
|---|---|---|
| TSB range over full history | −3.6 … +0.8 | −34.6 … +8.1 |
| Days reading fatigued | **0 of 72** | 22 orange, 1 red |
| `calcReadiness` `tsb < -20` fires | **0 days** | 8 days |
| `evaluateRules` R4 `tsb < -25` fires | **0 days** | 3 days |
| Peak CTL | 1.6 | 16.0 |

On the old scale both fatigue rules were **mathematically unreachable** — TSB could not get near their thresholds on any input, so the app was structurally incapable of ever reporting fatigue.

`sessionLoad` now emits classic TSS — `hours × sRPE × 10`, and `hours × IF² × 100` for the power path, which is exactly the standard TrainingPeaks formula. **Every existing threshold becomes correct untouched.**

A constant factor leaves #202's deliberate ~24% power-vs-sRPE offset intact, and a test now pins that ratio so a future rescale cannot quietly close it — that gap is read as fatigue in this project's own coaching notes, and calibrating it away on n=2 would erase the signal.

### 2. The wiring

`App.jsx` reads `useTSSHistory`. Both `DAILY_TSS` copies are gone — `constants/tssData.js` deleted, the byte-identical duplicate in `trainingConfig.js` removed.

## No stale fallback

Mobile used to fall back to that June array, so an outage rendered two-month-old numbers as current. Following `useAnchors`' documented rule and #196's precedent, a failed or empty read now says **TRAINING LOAD UNAVAILABLE** and the tiles show em-dashes — never a fabricated `0`, which reads as a confident "neutral".

Pending is deliberately **not** unavailable: a slow first read must not flash the outage line and then replace it with real numbers. Both outage tests were verified by reverting the guard and watching them go red.

## Also in scope

- **`MobileAnalytics`' recent-session list** kept a private pre-#202 sRPE-only copy of the formula, so a power-only session read `0` there while the chart directly above it counted it — two numbers for one session. It now calls `sessionLoad`.
- **`calcReadiness`** gained the explicit `tsb != null` guard its siblings `evaluateRules` and `autoregulate` already had.
- **`calcTrainingLoad` is unchanged** — it is scale-agnostic EWMA. Its callers now guard the empty series, which would otherwise walk from an Invalid Date.

## Verification

- CI green: 52 test files / 604 tests, coverage **lines 80.3 / branches 75.13 / functions 77.7** — above the ratchet, and slightly up on the recorded baseline.
- Numbers above produced by replaying the live session history through both scales.
- Cross-checked one row end to end: 2026-08-06, erg, 60:00, 152 W, no sRPE → `(152/205)² × 100 ≈ 55`.
- Live anchors used are `rowing_cp = 205`, `bike_ftp = 250` — read through `useAnchors`, not hardcoded.

**What this does not claim:** today's reading is TSB +8.1, still inside the neutral band. That is correct — TSB is bounded by CTL, and CTL has decayed to 10 after 16 days off. The fix is that the bands now apply to numbers on the scale they were designed for, not that today's verdict flips.

## Not in scope

S2b (vitals → `useVitals`) and S2c (strength PRs, `ftp = 190`, `totalErgDist`) are the next two PRs on #207. The `trainingConfig.js` half of the CP problem is #176, done in #203.

Part of #207.


skizzy1986:
**Fixed:** The desktop dashboard's CTL/ATL/TSB training-load numbers were being calculated on stale seed data and the wrong scale, so fatigue could never actually show up — those readings are now live and correctly banded, matching what mobile already showed. If training load data can't be loaded, the dashboard now clearly shows "TRAINING LOAD UNAVAILABLE" instead of quietly displaying old or fabricated numbers.

---

skizzy1986:
**Release note (Fixed):**

Fixed: the desktop dashboard's Fitness/Fatigue/Form (CTL/ATL/TSB) tiles and fatigue warnings were computed from a static, two-month-stale seed and calibrated to the wrong units — meaning fatigue could never actually be flagged, no matter how hard training got. These numbers now come from your live session history, on the correct scale, so the Form/readiness signals and fatigue warnings respond to real, current training load. If a reading can't be loaded, the dashboard now clearly shows "TRAINING LOAD UNAVAILABLE" instead of silently displaying old or fabricated numbers.


---

Issue #205, opened 2026-08-22
fix: LogSessionForm inserts sessions with no status, so they land NULL and count for nothing


skizzy1986:
Found during the #195 write-path audit. Pre-existing, latent, not introduced by that work.

## Problem

`sessions.status` is **nullable with no column default** (verified against the live DB: `column_default: null`, `is_nullable: YES`).

`web/src/components/LogSessionForm.jsx:72-81` inserts a session **without a `status` field at all**. So a strength session logged through the in-app form lands with `status = NULL`.

`NULL` is not in `COMPLETED_STATUSES` (`['actual', 'completed', 'logged']`), so such a row:

1. **Never enters `useTSSHistory`** — that query filters `.in('status', COMPLETED_STATUSES)`, so the session contributes nothing to CTL/ATL/TSB. The training-load numbers the dashboard leads with would silently omit it.
2. **Can never clear a benchmark badge** — `findMatch` in `web/src/utils/benchmarkStatus.js:63` gates on the same list and skips the row.

The session is in the database and renders in the log, but is invisible to every derived calculation.

## Why it hasn't bitten yet

All 92 live `sessions` rows have a non-null status, so the form has evidently never been the source of a saved row in practice. The defect is latent in the write path, waiting for the first real use.

## Fix

Decide the correct status value for a form-logged session (`'completed'` is the likely answer, matching the other bulk-import paths) and set it explicitly on insert.

Worth considering alongside: whether `sessions.status` should carry a NOT NULL constraint with a default, so this class of omission is rejected at the DB rather than silently accepted. That is a migration and a separate decision — additive and reversible, but it needs a backfill check first.

## Out of scope

Not fixed in #195's PR, which is confined to reconciling query invalidation across the three write paths. Fixing this means choosing a status value and possibly a migration — a behaviour change, not an invalidation fix.

PR #212, opened 2026-08-22
fix: make all three session-insert paths invalidate the same queries (#195)

### LINKED ISSUE #195: Offline queue drain invalidates no queries

skizzy1986:
Closes #195

## What changed and why

The offline queue drained on reconnect and invalidated nothing, so a session saved offline stayed invisible to every view until a `staleTime` lapsed or the app remounted. Auditing the other write paths showed the real problem is that all three disagreed — and none invalidated `['tss-history']`, which nothing in the codebase has ever invalidated, so CTL/ATL/TSB could be stale for up to five minutes after a save.

| Path | Before | After |
|---|---|---|
| `LogSessionForm.jsx` | `['sessions']` | all three |
| `ErgLiveView.jsx` | `['sessions']`, `['erg-sessions']` | all three |
| `useOfflineQueue.js` drain | **nothing** | all three |

`utils/invalidateSessionQueries.js` is now the single definition of what a session insert must refresh. A helper rather than three aligned copies because this drift has already happened twice — once for `['erg-sessions']`, once for `['tss-history']`.

**Correction to the issue body:** it proposed invalidating `['benchmark-sessions']`. No such key exists. The benchmark window lives at `['sessions','benchmark-window']`, a prefix-child of `['sessions']`, so it was already being reached. One test pins that the helper does not list it separately; another proves the cascade behaviourally by seeding a query and asserting it goes stale.

The drain needed no plumbing. `drainQueue` is module-private with one caller — the `sync()` closure, which already runs under `QueryClientProvider` and already guards on `synced > 0`. Invalidating once inside that guard gives correct partial-drain behaviour for free: some rows landed, refresh with what did; nothing landed, invalidate nothing, matching `LogSessionForm`'s existing failure contract. No signature change, no singleton (that would create an import cycle through `main.jsx`).

## How to test manually

Covered by CI. For a live check: go offline, save a session from the PM5 live view, reconnect — the session list, erg list and training-load numbers should all refresh without a reload.

## Checklist

603 tests (+8). Eight existing `renderHook` sites in `useOfflineQueue.test.js` gained a `QueryClientProvider`, which the hook now requires.

**Merge order:** this one first — it is clean against both other open PRs.

Out of scope: `useSessionLog.js` (#194), the NULL-`status` bug in `LogSessionForm` (#205).

---

### DIFF
--- a/web/src/components/LogSessionForm.jsx
+++ b/web/src/components/LogSessionForm.jsx
@@ -3,6 +3,7 @@ import { useMutation, useQueryClient } from '@tanstack/react-query';
 import { supabase } from '../supabaseClient.js';
 import { THEME } from '../constants/theme.js';
 import { toLogDate } from '../utils/dateFormat.js';
+import { invalidateSessionQueries } from '../utils/invalidateSessionQueries.js';
 
 export default function LogSessionForm({ onSaved }) {
   const queryClient = useQueryClient();
@@ -82,7 +83,7 @@ export default function LogSessionForm({ onSaved }) {
       if (error) throw error;
     },
     onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: ['sessions'] });
+      invalidateSessionQueries(queryClient);
       setMsg({ type: 'ok', text: 'Saved! Session added to your log.' });
       reset();
       if (onSaved) onSaved();
--- a/web/src/components/__tests__/LogSessionForm.test.jsx
+++ b/web/src/components/__tests__/LogSessionForm.test.jsx
@@ -83,6 +83,8 @@ describe('LogSessionForm', () => {
 
     await waitFor(() => expect(insertMock).toHaveBeenCalledTimes(1));
     expect(invalidateSpy).toHaveBeenCalledWith({ queryKey: ['sessions'] });
+    expect(invalidateSpy).toHaveBeenCalledWith({ queryKey: ['erg-sessions'] });
+    expect(invalidateSpy).toHaveBeenCalledWith({ queryKey: ['tss-history'] });
     expect(onSaved).toHaveBeenCalledTimes(1);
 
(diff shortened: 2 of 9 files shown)


skizzy1986:
**Fixed** — Sessions saved while offline (or logged from the live PM5 erg view) sometimes didn't show up in your session list, erg history, or training-load (CTL/ATL/TSB) numbers right away — you'd have to wait several minutes or reload the app to see them. All three save paths now refresh the same data consistently, so a session you log — online or off — appears everywhere immediately once it's saved.

---

Issue #206, opened 2026-08-22
useErgSessions filters status='logged', which no session row matches — ErgView's list is empty


skizzy1986:
Found during the #200 research. Pre-existing, not introduced there.

## Observation

`web/src/hooks/useErgSessions.js` filters `.eq('type', 'erg').eq('status', 'logged')`.

Against the live DB today (2026-08-22), that returns **zero rows**:

| type | status | rows |
|---|---|---|
| erg | completed | 18 |
| erg | cancelled | 11 |
| erg | actual | 5 |
| erg | **logged** | **0** |

34 rows are `type='erg'`; none has `status='logged'`.

## Why

`'logged'` is written only by the live PM5 Bluetooth save path (`web/src/views/ErgLiveView.jsx:641` writes `status: 'logged'`), per the status convention in `CLAUDE.md`. Every erg row currently in the table is bulk-imported history carrying `completed` or `actual`.

So the filter is doing exactly what it was written to do — there simply has never been a PM5-saved session. `ErgView`'s session list renders empty and will stay empty until the first live save.

## The question

Is that the intended behaviour?

- **If yes** — this is just an empty state, and the issue is only whether it reads as "no sessions yet" rather than looking broken.
- **If no** — the hook should widen to `COMPLETED_STATUSES` (or `.in('status', [...])`) so imported erg history appears alongside live-logged sessions.

This is a product decision about what ErgView is for, not a defect with an obvious fix, which is why it is filed rather than folded into #200.

## Note

#200 moves this hook onto `date_iso` ordering. That change is correct but, given the empty result set, currently a practical no-op — the ordering only starts mattering once rows actually match the filter. #200 deliberately does **not** widen the status filter.

skizzy1986:
**Triage**

- **Priority: Medium** (matches the `P2` label). Not a crash — `useErgSessions` is filtering exactly as written, and the empty result is an empty-state UX question rather than a defect with an obvious fix. Worth a product decision (widen to `COMPLETED_STATUSES` or accept the empty state) before it's mistaken for broken.
- **Duplicate check:** none found. Related to #207 (S2 — Data truth) but distinct scope — #207 doesn't touch `useErgSessions`'s status filter.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #213, opened 2026-08-22
chore: order useErgSessions and useBenchmarkSessions by date_iso (#200)

### LINKED ISSUE #200: chore: move useErgSessions and useBenchmarkSessions onto sessions.date_iso
Follow-up to #187 / #198, which added a generated `date_iso` date column to `sessions` (indexed on `(user_id, date_iso desc nulls last)`) and moved `useSessions` and `useTSSHistory` onto it. Two hooks were deliberately left on the old text ordering.

**`web/src/hooks/useErgSessions.js`** — still `.order('date', {ascending: false}).limit(50)`, corrected afterwards by a client-side `useMemo` re-sort. Its comment says the `.limit(50)` still fetches the wrong 50 and names #187 as the open root cause. That is still literally true while it orders on `date`, but #187 is now closed, so the reference is stale and the real fix is available.

**`web/src/hooks/useBenchmarkSessions.js`** — still `.order('date', {ascending: false}).limit(500)`. The 500 was chosen specifically because "the server-side order is not chronological and a small limit truncates the wrong rows". `date_iso` removes that rationale. Keep the wide limit if the ladder resolver genuinely wants full history, but say so in the comment rather than citing a sort workaround that no longer applies.

Per hook this is a one-line change:

```js
.order('date_iso', { ascending: false, nullsFirst: false })
```

`nullsFirst: false` is **not optional** — `postgrest-js` omits the nulls clause entirely when the option is undefined, and Postgres defaults descending to NULLS FIRST, which would float an unparseable date to the top as the most recent row.

Also worth deciding here: whether to keep the client-side re-sorts as defence (they cover the non-atomic migrate-then-deploy window and any rollback, and they are what keeps the hooks' ordering tests meaningful) or drop them as redundant. #198 kept the equivalent guard in `useSessions` for exactly those reasons.

Rewrite both comments in the same PR — a comment describing a bug that no longer exists is worse than no comment.

skizzy1986:
Closes #200

## What changed and why

#187/#198 added the generated `date_iso` column and moved `useSessions` and `useTSSHistory` onto it. These two hooks were left on the lexical TEXT ordering. Both now order by `date_iso` with `nullsFirst: false` — not optional, since postgrest-js omits the clause entirely when the option is undefined and Postgres defaults descending to NULLS FIRST, which would float an unparseable date to the top as the most recent row.

| | `useErgSessions` | `useBenchmarkSessions` |
|---|---|---|
| `date_iso` + `nullsFirst: false` | ✅ | ✅ |
| `id` tiebreaker | ✅ — `limit(50)` is a real cut and the PM5 path suffixes labels with `hh:mm`, so two sessions can share a date | ❌ deliberate — `limit(500)` against 92 rows never binds, and the resolver already tiebreaks by id itself |
| client-side re-sort | kept | none, and none added |

The asymmetry is pinned by a test asserting `useBenchmarkSessions` does *not* order by `id`, so a future "consistency" cleanup fails loudly.

**Comment corrections.** The stale comment cited #184, not #187 as the issue body claimed; both rewritten comments now cite #187, the issue that explains why `date_iso` exists. A third stale comment in `benchmarkStatus.js` asserted "the payload is ordered by the TEXT date column" — the premise this PR falsifies — and is now past tense. A fourth was corrected in review: the re-sort's rationale claimed it survives a rollback or a stale PostgREST schema cache, which it cannot (a 400 leaves no payload to re-sort). What it actually guards is an unparseable date arriving NULL, which is what the new test covers.

## How to test manually

No user-visible change. `useBenchmarkSessions` returns the whole table either way at 92 rows; `useErgSessions` returns an empty set in production because no row carries `status='logged'` yet (see #206). This is correctness insurance for data the table hasn't reached.

## Checklist

607 tests (+12), including a new `useBenchmarkSessions.test.js` — it was the only one of the four session hooks with no direct test, and this PR changes its query contract.

**Merge order:** second, after #195. Conflicts with the #192 PR on one comment block in `benchmarkStatus.js`; #192 rebases onto this.

Out of scope: widening `useErgSessions`' `status='logged'` filter (#206), any limit, select list, query key or `staleTime`.

---

### DIFF
--- a/web/src/hooks/__tests__/useBenchmarkSessions.test.js
+++ b/web/src/hooks/__tests__/useBenchmarkSessions.test.js
@@ -1,0 +1,117 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderHook, waitFor } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import React from 'react';
+
+const fromMock = vi.fn();
+const orderMock = vi.fn();
+const limitMock = vi.fn();
+
+vi.mock('../../supabaseClient.js', () => ({
+  supabase: {
+    from: (...args) => fromMock(...args),
+  },
+}));
+
+import { useBenchmarkSessions } from '../useBenchmarkSessions.js';
+
+function makeClient() {
+  return new QueryClient({ defaultOptions: { queries: { retry: false } } });
+}
+
+function makeWrapper(client) {
+  function Wrapper({ children }) {
+    return React.createElement(QueryClientProvider, { client }, children);
+  }
+  return Wrapper;
+}
+
+// order() spies on its arguments and keeps returning the chain; limit()
+// terminates it. The query contract is what this hook is — there is no derived
+// state to assert on.
+function mockQuery(data, error = null) {
+  const chain = {
+    select: () => chain,
+    order: (...args) => {
+      orderMock(...args);
+      return chain;
+    },
+    limit: (...args) => {
+      limitMock(...args);
+      return Promise.resolve({ data, error });
+    },
+  };
+  fromMock.mockReturnValue(chain);
+}
+
+async function renderBenchmarkSessions(client = makeClient()) {
+  const { result } = renderHook(() => useBenchmarkSessions(), {
+    wrapper: makeWrapper(client),
+  });
+  await waitFor(() => expect(result.current.isSuccess).toBe(true));
+  return result;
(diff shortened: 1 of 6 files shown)


skizzy1986:
Internal change — no release note. This fixes server-side query ordering (`date_iso` instead of the lexical `date` text column) in two data hooks plus adds test coverage; per the PR description it has no user-visible effect today.

---

PR #214, opened 2026-08-22
feat: show rescheduled benchmarks instead of warning forever (#192)

### LINKED ISSUE #192: Benchmark reschedule/acknowledge state (T2 follow-up)

skizzy1986:
Closes #192

## What changed and why

CP Test #1, CP Test #2 and the 5k are overdue because they never happened, so they render `OVERDUE` indefinitely — roughly 197 days by the time the Jan 2027 windows open. A banner showing the same three warnings for six months becomes wallpaper, and alarm fatigue is how the original silent-slip bug recurs.

Rescheduling is expressed by putting a `planned` session on the calendar, not by a dismiss button. That capability already existed; the badge just never looked at it. **This is a read change** — no writes, no schema change, no new hook, nothing device-local that Coach cannot see.

An overdue benchmark with a future `planned` session that names it now renders `↻ RESCHEDULED · 9/12/26 · 31d OVERDUE`. Nothing goes silent: the row still renders, still names the benchmark, still carries `daysOverdue`.

**Containment is structural, not trusted.** A second pass runs after the existing matcher, iterates only states already resolved to `overdue`, and keeps its own claim set. So `scheduled` has exactly one inbound edge and a `planned` row has no path to `done` — verifiable by reading eight lines. The two status gates stay in two separate functions and the status list is never parameterised; only the label predicate is shared, so the ride-elevation exclusion cannot regress in one pass alone.

A `planned` row dated before today is a missed plan, not a reschedule, and leaves the warning standing — live sessions 43 and 56 are exactly that case, and there's a test replaying them.

## ⚠️ What you will and won't see on merge

**The reschedule half ships dormant.** No future-dated, label-matching `planned` row exists, so **no badge changes today**. Validation confirmed this by running the resolver against the live 92-row payload.

**The sort half ships live.** The panel reorders on first render:

| | Before | After |
|---|---|---|
| 1 | CP Test #1 *(no badge)* | **CP Test #2** `OVERDUE 33d` |
| 2 | CP Test #2 `OVERDUE 33d` | **5k Time Trial** `OVERDUE 12d` |
| 3 | 5k Time Trial `OVERDUE 12d` | CP Test #1 |
| 4–5 | Erg Power Series / C2 Monthly | *unchanged* |

Please don't file "rescheduling doesn't work" — it works, it's waiting for its first input.

## How to test manually

Insert a `planned` session dated in the future with a label naming a benchmark (e.g. `CP RETEST — 1min + 4min max`) and the CP Test #2 row should flip to the reschedule badge and sort below the remaining overdue row.

## Checklist

620 tests (+25). Mutation-checked: disabling the second pass fails exactly 7 tests; reversing slice/sort fails exactly 1; folding the `scheduled` branch back into the trailing `else` fails 4.

**Merge order:** last. Rebase onto #213 first — they conflict on one comment block in `benchmarkStatus.js` (keep this branch's `labelMatches` **and** #213's past-tense wording; the code below the marker is byte-identical on both sides).

**Known limitation, filed separately:** the visible five are still chosen by ladder position before severity sorting, so a benchmark at index ≥ 5 going overdue stays invisible. Bites Feb 2027.

---

### DIFF
--- a/web/src/components/BenchmarkBadge.jsx
+++ b/web/src/components/BenchmarkBadge.jsx
@@ -1,11 +1,18 @@
+import { toLogDate } from '../utils/dateFormat.js';
+
 // Presentational only — props in, no hooks, no data access.
-const COLOURS = { overdue: '#ff2d55', upcoming: '#ffd700' };
+const COLOURS = {
+  overdue: '#ff2d55',
+  upcoming: '#ffd700',
+  scheduled: '#a78bfa',
+};
 
 export default function BenchmarkBadge({
   status,
   fuzzy = false,
   daysOverdue = null,
   daysUntilStart = null,
+  rescheduledTo = null,
 }) {
   const colour = COLOURS[status];
   if (!colour) return null;
@@ -15,6 +22,14 @@ export default function BenchmarkBadge({
     parts.push('OVERDUE');
     if (typeof daysOverdue === 'number' && daysOverdue > 0) {
       parts.push(`${daysOverdue}d`);
+    }
+  } else if (status === 'scheduled') {
+    // Explicitly branched, never folded into the trailing else: a scheduled
+    // benchmark falling through to the upcoming branch would read 'DUE'.
+    parts.push('↻ RESCHEDULED');
+    if (rescheduledTo) parts.push(toLogDate(rescheduledTo));
+    if (typeof daysOverdue === 'number' && daysOverdue > 0) {
+      parts.push(`${daysOverdue}d OVERDUE`);
     }
   } else {
     parts.push('DUE');
--- a/web/src/components/__tests__/BenchmarkBadge.test.jsx
+++ b/web/src/components/__tests__/BenchmarkBadge.test.jsx
@@ -53,4 +53,54 @@ describe('BenchmarkBadge', () => {
     rerender(<BenchmarkBadge status="upcoming" daysUntilStart={3} />);
     expect(container.firstChild).toHaveStyle({ color: '#ffd700' });
   });
+
+  // AC4 — the rescheduled state has to be legible as its own thing: it is not
+  // an alarm (the date is booked) and it is not routine (it is still late).
+  it('AC4 renders the reschedule date and how late the benchmark still is', () => {
+    render(
+      <BenchmarkBadge
+        status="scheduled"
+        rescheduledTo="2026-09-12"
+        daysOverdue={31}
+      />
+    );
+    expect(
+      screen.getByText('↻ RESCHEDULED · 9/12/26 · 31d OVERDUE')
+    ).toBeInTheDocument();
+  });
+
+  it('AC4 colours scheduled violet, distinct from overdue and upcoming', () => {
+    const { container } = render(
+      <BenchmarkBadge status="scheduled" rescheduledTo="2026-09-12" />
+    );
+    expect(container.firstChild).toHaveStyle({ color: '#a78bfa' });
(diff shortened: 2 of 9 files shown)


skizzy1986:
**Release notes draft (Improved)**

Overdue benchmark badges on the season ladder no longer stay stuck on "OVERDUE" forever once a retest is back on the calendar — if a future session is scheduled that names the benchmark, its badge now switches to "↻ RESCHEDULED" with the new date, while still showing how many days it's been outstanding. The ladder panel also now sorts its five visible rows so overdue and due-soon benchmarks float to the top, instead of always following fixed ladder order.

---

PR #218, opened 2026-08-22
fix: put recovery on live vitals, and close the readiness fork

skizzy1986:
Second of three PRs for **S2 — Data truth** (#207). Follows #208.

## The problem

`RecoveryView` and the home screen scored readiness from `recoveryLog` — a static constant whose last entry is **6/20** — against two baselines hand-set in June. Mobile read the same athlete's live vitals. Two screens, one question, different answers.

The fork wasn't just the data source. There were **two readiness functions**:

| | `calcReadiness` (`analysis.js`) | `computeReadiness` (`recoveryAnalytics.js`) |
|---|---|---|
| Baselines | static 57 / 30 | live rolling personal |
| TSB term | yes | **no** |
| Bands | 75 / 50 | 80 / 60 |
| Labels | READY / CAUTION / **REST** | READY / CAUTION / **FATIGUED** |
| No-data guard | explicit `NO DATA` | **none** |

They're now one function in `recoveryAnalytics.js`: live baselines from that side, the TSB term and the NO DATA guard from this one. The brief's Today hero is a readiness score, so S4 needs a single definition to build on.

## Three real bugs this surfaced

**1. Absent data produced a confident verdict — in both directions.** `computeReadiness` scored a missing row `0 / FATIGUED` and a row whose metrics were all null `100 / READY`. Both invented from nothing. The `0` also defeated `CoachView`'s `readinessScore != null` guard, so a fabricated readiness reached the **Coach prompt as fact**. Missing RHR now yields `score: null` / `NO DATA`.

**2. Baseline provenance was reported per set, not per metric.** `hasPersonalBaselines` was `rows.length >= 14` — a count of rows, not of usable readings. In the live data **17 of the last 30 days carry RHR but only 6 carry HRV**, so the flag read true while the HRV baseline was still the population default, and the view labelled `30` *"your avg"*. `computePersonalBaselines` now reports `rhrPersonal` / `hrvPersonal` separately and each metric is labelled on its own.

**3. `evaluateRules`' R4 compared the athlete to June.** It read the two static constants, so "HRV below baseline AND RHR above baseline" measured today against a hand-set number rather than against the athlete. It now takes `baselines`, defaulting to the population numbers.

## Decisions taken

Two judgement calls, flagged because they're reversible and yours:

- **Bands kept at 80/60 with FATIGUED** (mobile's, not desktop's 75/50 + REST). With live baselines the reference is the athlete's own rolling trimmed mean, so day-to-day deviations are small and scores cluster high — looser bands would read READY nearly every day. FATIGUED also names a *state*, leaving the prescription to `autoregulate`'s GREEN/AMBER/RED. `autoregulate`'s own 50/75 cutoffs are deliberately **untouched**: it reads the raw score, not the label, so the band change doesn't reach it.
- **The no-data guard is folded in here** rather than filed separately, because it's the same bug class as #208's and it was actively feeding the Coach.

## Also

`useVitals` sources TSB internally via `useTSSHistory` (shared query key, served from cache) so there is exactly one readiness number per day and the two surfaces cannot drift apart again.

`recoveryLog`, `RHR_BASELINE` and `HRV_BASELINE` are deleted — `vitals` holds 2026-03-23 onward, a superset of the constant. `App.jsx`'s baseline import was already dead after #203 removed its siblings.

## Verification

- Full suite green locally: **52 files / 610 tests**; `npm run lint` 0 errors; `npm run build` clean.
- The defaulted-baseline label test was verified by reverting the fix and watching it go red.
- Checked against live data: latest vitals row is 2026-08-21 (RHR 57, HRV 32, sleep 5.9), so the screen has something real to score.

## Not in scope

S2c (strength PRs → `useStrengthPRs`, the surviving `ftp = 190`, `totalErgDist`) is the last PR on #207.

Part of #207.


skizzy1986:
## Release notes draft

**Fixed — Readiness score now consistent and accurate**

Your readiness score on the Today screen and the Recovery view is now calculated the same way, using your live rolling vitals instead of a stale baseline from June. Previously, a day with missing recovery data (like no resting heart rate) could silently show a score of 0 or 100 instead of "no data" — and that fabricated number could even reach the Coach's recommendations. Baseline labels (e.g. "your avg" HRV) now only display once there's genuinely enough history behind them, rather than being triggered by unrelated heart-rate data.


---

PR #220, opened 2026-08-22
docs: record the design workstream state of play

skizzy1986:
Adds `web/.design-sync/STATE_OF_PLAY.md` — a status companion to `DESIGN_BRIEF.md`, put in the repo so a Claude Design session can read it directly rather than from a chat transcript.

Rendered version: https://claude.ai/code/artifact/e16750c2-4d67-4bc2-99a3-7fae8ca450dc

## Why this is worth a file

**The claude.ai/design project has never been opened.** The `_ds_needs_recompile` sentinel is still the one the upload wrote (`{"by":"design-sync-cli"}`) and there is no `_ds_manifest.json` — the app generates that on first open and clears the sentinel. So this is positive evidence, not absence of change.

**The design-sync build is broken on `main`.** #203 correctly removed the static `PACE_ZONES` export from `trainingConfig.js` so a seed-derived zone table cannot diverge from the live CP anchor — but `web/.design-sync/entry.jsx:20` still re-exports the symbol. Reproduced against the exact `origin/main` sources:

```
x No matching export in "src/constants/trainingConfig.js" for import "PACE_ZONES"
```

This is the first entry in `NOTES.md`'s *Re-sync risks* — "a hand-maintained barrel with no drift detection" — coming true within a day of being written. The note wasn't sufficient; the barrel wants a CI guard.

**`conventions.md` ships a WCAG AA failure.** It recommends `THEME.muted` on `surface`/`raised` (4.34:1 and 3.50:1). That file is inlined into the design agent's prompt, so the failure propagates into every generated design. `textSubtle` clears it at 6.13:1. This is §1.6 of the brief, re-verified.

## Also recorded

Three of `DESIGN_BRIEF.md` §8's open questions have closed: the typeface (IBM Plex Sans + Plex Mono, decided on metric compatibility), the styling seam (approved — `style` merged last on layout primitives only), and `rowing_cp` — which #203 resolved a third way neither option in §8.3 anticipated, by deleting the hardcoded mirror entirely.

## Scope

Documentation only. **Neither defect is fixed here** — both are queued in §5 of the document, and the barrel repair plus the contrast fix should land together in their own PR followed by a re-sync.

### DIFF
--- a/web/.design-sync/STATE_OF_PLAY.md
+++ b/web/.design-sync/STATE_OF_PLAY.md
@@ -1,0 +1,154 @@
+# SplitIQ — design workstream, state of play
+
+_22 August 2026. Companion to [`DESIGN_BRIEF.md`](DESIGN_BRIEF.md) (the prescriptive
+document) and [`conventions.md`](conventions.md) (the descriptive one). This file is
+neither — it is a status check, and it goes stale quickly._
+
+Rendered version: https://claude.ai/code/artifact/e16750c2-4d67-4bc2-99a3-7fae8ca450dc
+
+> **Verdict.** The Claude Design project has never been opened. Separately, `main`
+> moved underneath the synced design system and the design-sync build is now
+> **broken on `main`** — verified by reproduction, not inferred from a diff.
+
+---
+
+## 1. The design project has never been opened
+
+The upload's final step writes `_ds_needs_recompile`. The claude.ai/design app clears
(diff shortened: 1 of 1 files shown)


skizzy1986:
Internal change — no release note. This adds `web/.design-sync/STATE_OF_PLAY.md`, an internal status doc for the design-sync workstream; no user-facing behavior changed.

---

PR #224, opened 2026-08-22
fix: repair the design-sync barrel and the contrast it was teaching

skizzy1986:
Fixes both defects recorded in `STATE_OF_PLAY.md` §2 (#220). Both are ours, both are now re-uploaded to the [design project](https://claude.ai/design/p/8edb6124-729f-4d6c-9e32-6ba955e2a6b4).

## 1. The barrel re-exported a symbol that no longer exists

#203 correctly deleted the static `PACE_ZONES` export from `trainingConfig.js` so a seed-derived zone table couldn't diverge from the live `rowing_cp` anchor. `entry.jsx` wasn't updated, so the design-sync build failed outright:

```
x No matching export in "src/constants/trainingConfig.js" for import "PACE_ZONES"
```

That blocked **every** re-sync. `entry.jsx` now exports `derivePaceZones`, and the `PaceTrendChart` preview calls it with the current CP anchor. `docs/`, `config.json` and `conventions.md` are updated to match, so the design agent is told to *derive* bands rather than import a table that no longer exists.

## 2. `conventions.md` was teaching a WCAG AA failure

It recommended `THEME.muted` at `fontSize: 9` on `surface`/`raised` panels — 4.34:1 and 3.50:1, both under the 4.5:1 threshold (`DESIGN_BRIEF.md` §1.6). That file is inlined into the design agent's prompt, so the failure reproduced in **every** design it generated.

Fixed at source — and the **rule** was added, not just the instance. `conventions.md` now carries the full contrast table and states plainly that `textSubtle` is the safe choice on `surface`/`raised`, `muted` is only safe on `bg`, and `textFaint`/`textDim` pass nowhere and are decorative-only. A corrected example alone would have left the agent free to reach for `muted` elsewhere.

## What NOTES.md now records

The barrel incident, including the uncomfortable part: the *existing* "hand-maintained, no drift detection" warning **did not prevent it**. A prose note is not a control. The durable fix is a CI step that bundles `entry.jsx` on any `web/src` change — a two-second esbuild run that catches the whole class. Filed as a follow-up rather than smuggled into this PR.

Also records the two couplings from the approved brief decisions: `dtsPropsFor` must document the `style` prop when the layout primitives land (or the agent codes against a contract denying the seam exists), and `cfg.extraFonts` must carry the IBM Plex woff2 files once self-hosted (or generated designs keep rendering in fallback after the app is fixed).

## Verification

- `package-build.mjs` and `package-validate.mjs` both exit 0; render check **11/11 clean**.
- All 11 components re-graded from fresh capture sheets — every grade key moved because global config changed, so none were carried over blind.
- Closing driver run reports `pendingGrade: []` and `deletePaths: []`.
- Uploaded in the documented order: sentinel fence → 61 content files → sentinel re-arm → `_ds_sync.json` last.

Sync inputs only — no application code touched.


skizzy1986:
Internal change — no release note. All changes are confined to `web/.design-sync/` (the design-agent sync barrel, docs, and conventions) — no application code was touched, as the PR description itself confirms ("Sync inputs only").

---

PR #225, opened 2026-08-22
ci: guard the design-sync barrel against silent drift

skizzy1986:
The follow-up promised in #224. **Based on `fix/design-sync-barrel-and-contrast`, not `main`** — the guard would fail on `main`, where the barrel is still broken. Merge #224 first and this retargets cleanly.

## Why

`web/.design-sync/entry.jsx` re-exports app components by name so the claude.ai/design bundle can expose them on `window.SplitIQ`. **Nothing else in the repo imports it**, so renaming, moving or deleting anything it names breaks the design-sync build with no other symptom — invisibly, until someone next runs a sync.

That already happened: #203 correctly removed the static `PACE_ZONES` export from `trainingConfig.js`, the barrel wasn't updated, and the break sat on `main` undetected. `NOTES.md` carried a "hand-maintained, no drift detection" warning at the time — and it did not prevent it. A prose note is not a control.

## What it does

`npm run check:design-sync` bundles the barrel and fails if any re-export no longer resolves. It also checks every `cfg.componentSrcMap` path still exists, catching a component file that *moved* rather than an export that vanished.

Runs in the **Lint & Format** job, which already does `npm ci` and is gated by the existing paths-filter on `web/**` — so it fires on exactly the changes that can break it, at no extra install cost.

## Why rolldown, not esbuild

Vite 8 ships **rolldown**, so it's already in the tree — no dependency added for a CI check. (There is no esbuild in `web/node_modules` at all; the converter installs its own into the gitignored `.ds-sync/`.) A missing named re-export is a hard error in either bundler, which is the class being guarded.

## Verified both ways

Reintroducing the #203 regression:

```
design-sync entry guard FAILED

  ✗ .design-sync/entry.jsx does not bundle:

[MISSING_EXPORT] "PACE_ZONES" is not exported by "src/constants/trainingConfig.js".
    ╭─[ .design-sync/entry.jsx:23:10 ]
 23 │ export { PACE_ZONES } from '../src/constants/trainingConfig.js';
    │          ─────┬────
    │               ╰────── Missing export
```

exit 1. A moved `componentSrcMap` path is reported by name. The repaired barrel exits 0.

## Note

`scripts/` is outside `npm run lint` / `format:check` scope (both are `src/`-only), matching the existing `check-bundle-size.mjs`. The new script is prettier-formatted anyway and has no ESLint errors beyond the Node-globals ones the existing scripts also carry.


skizzy1986:
Internal change — no release note (adds a CI check that guards the design-sync barrel against silent drift; no user-visible behavior change).

---

PR #226, opened 2026-08-22
feat: give EVENT_LADDER benchmarks a real link to their session (#188)

### LINKED ISSUE #188: feat: give EVENT_LADDER benchmarks a real link to their session (replace keyword matching)
Follow-up to #175 (Sprint 5 · T2). Approved as a separate issue at that feature's Gate 2 — the durable fix for the one weak point in the benchmark due/overdue indicator.

## The problem

`EVENT_LADDER` (`web/src/constants/schedule.js:313-386`) entries carry exactly five fields — `date`, `name`, `kind`, `phase`, `serves`. **There is no link of any kind to a `sessions` row**: no session id, no label key, no date join. Nothing.

So #175 has to answer "was this benchmark actually done?" heuristically. It derives keywords from `entry.name` and matches them as whole tokens against `sessions.label`, inside a date window with a 14-day backward grace:

| Entry name | Derived keywords |
|---|---|
| `CP Test #1 (4-min)` | `['cp', '4min']` |
| `CP Test #2 (2nd duration)` | `['cp']` |
| `5k Time Trial` | `['5k']` |
| `2k Test` | `['2k']` |
| `1000m + 1-min tune-ups` | `['1000m', '1min']` |

This works correctly on today's data and every case is tested — but it is matching prose against prose, and it fails in both directions:

- **False done** — a routine session whose label happens to contain a keyword token clears the benchmark. A `2k warm-up paddle` logged during the 2k Test window would mark the 2k test complete.
- **False overdue** — a benchmark that *was* done but logged with different vocabulary never clears. `Critical Power retest` contains no `cp` token and no digit token, so it would nag forever.
- **Cross-entry collision** — CP Test #1 and #2 share the keyword `cp` and their grace ranges overlap. #175 handles this with a claim-once rule (a session that clears one benchmark cannot clear another), which is a workaround for the missing link, not a fix.

#175 deliberately biases toward false-overdue (a nag beats a silently missed test) and documents each failure mode with a test that asserts the known-wrong behaviour. Those tests are the tripwire for this issue.

## Suggested direction

Give the pairing something structural to key off. Two shapes worth weighing at spec time:

1. **A `benchmark_key` on both sides** — a stable slug (`cp-test-1`, `5k-tt-base-end`) as a new `EVENT_LADDER` field and a new nullable column on `sessions`. Additive and reversible; lets a session declare which benchmark it satisfies at log time.
2. **A `session_id` on the ladder entry** — simpler, but the ladder is static constants and the session may not exist yet when the entry is authored, so it inverts awkwardly.

(1) looks stronger, but this needs a real spec — note that restructuring `EVENT_LADDER` was explicitly out of scope for #175, so this is the issue where that question gets decided.

## What this would replace

Once a structural link exists, `web/src/utils/benchmarkStatus.js` loses `benchmarkKeywords`, `BENCHMARK_TERMS`, the whole-token matching, the grace window, and the claim-once set — the matcher collapses to a lookup. The window parser (`web/src/utils/eventWindow.js`) stays; it is still needed for the fuzzy `~Mid Jul 26` dates and for overdue/upcoming timing.

Also worth revisiting here: #178 (T5, pre-benchmark protocol checklist) consumes `selectUpcoming()` from the same module, so its seam should survive this change unchanged.

skizzy1986:
> ## 🔄 Rebased onto `main` (`9560364`) — read this first
>
> This branch was cut at `c0acf7d`. Two commits landed on `main` afterwards that touch the same resolver, and reconciling them **narrows one claim made below**.
>
> **`#192`/`#214` — rescheduled benchmarks.** `main` added a *second* pass that reads **PLANNED** rows by label keyword and can move a badge `overdue → scheduled`. It is preserved intact, so `benchmarkKeywords`, `BENCHMARK_TERMS`, `tokenize` and `labelMatches` **survive in `eventWindow.js`** rather than being deleted as this PR originally had them.
>
> The two features are genuinely independent, and the split is now the design:
>
> | | reads | can set | fires when |
> |---|---|---|---|
> | Pass 1 — **done** | **completed** rows, by `benchmark_key` | `done`, `quiet` | always |
> | Pass 2 — **reschedule** | **planned** rows, by keyword | `scheduled` | only if already `overdue` |
>
> Disjoint row sets, separate claim sets, and pass 2 never touches a `done` benchmark. A planned row still cannot fabricate a completion — `main`'s own comment puts it best: *"A planned row is a commitment, never evidence."*
>
> **So the honest invariant is narrower than "Label text is never read" below:** label text is never read to **clear** a benchmark, only to **reschedule** one. Keyword matching is safe in pass 2 precisely because its worst case is a badge reading "moved" instead of "late".
>
> **`#200`/`#213` — `date_iso` ordering.** Kept, alongside `benchmark_key` in the select. The "Deliberately left alone" note about `.order('date', …)` below is **resolved**, not outstanding.
>
> **Two tests changed meaning.** `exposes no keywords field on any state` became `derives keywords only for benchmark entries` (the field is needed again); and the fixture in `leaves a done benchmark quiet, planned session or not` now establishes done-ness **by link** rather than by label — which is the whole point of this change. The `CANCELLED_RETEST` and `EARLY_AUG_DONE` fixtures are restored for the live-ladder assertion.
>
> **Why this PR had no web CI until now:** `ci-web.yml` never triggered on the original push — only `ci-android` and the PR-title lint ran, so Lint, Test, Build, Playwright and CodeQL had **never** executed against these 18 files. The rebase-push fixed that; the results below are the first real verdict.

---

## What this changes

`EVENT_LADDER` benchmarks had no link of any kind to a `sessions` row, so
"was this benchmark actually done?" was answered heuristically: derive keywords
from the benchmark's name (`CP Test #1 (4-min)` → `['cp','4min']`), match them as
whole tokens against `sessions.label` inside a date window with 14 days of
backward grace, and add a claim-once rule so one session could not clear two
benchmarks.

That matched prose against prose, and failed in both directions — a routine
session whose label happened to contain a keyword cleared the benchmark, and a
benchmark genuinely done but logged with different words (`Critical Power
retest`) nagged forever.

**A benchmark is now done only when a completed session carries its slug in the
new `sessions.benchmark_key` column. Label text is never read to clear one.**

## Design decisions worth reviewing

- **A slug on both sides, not a `session_id` on the ladder.** `EVENT_LADDER` is a
  git-committed constant shipped in the Vercel bundle, so a session id there would
  make every link a code change → PR → CI → deploy. Worse, Coach-over-MCP could
  not create a link at all, since Coach holds DB access, not write access to a
  deployed bundle.
- **The 5 slugs are hand-written literals, never derived from `name`.** Deriving a
  slug means a ladder rename silently orphans the link. A test asserts the five
  literal values, and that non-benchmark entries have no `key` field at all.
- **Dates no longer gate "done".** A date check would be a second implicit input to
  the done decision — the same bug class this PR removes, in a different costume. A
  test logged three weeks after a slipped `~Mid Jul` window still counts. The window
  stays fully live for timing **unlinked** benchmarks (overdue / upcoming / fuzzy).
- **Uniqueness is held at the database**, by a partial unique index
  `(user_id, benchmark_key) WHERE benchmark_key IS NOT NULL`, mirroring
  `anchors_one_live_per_key`. That is the half that survives a Coach-over-MCP write
  with no app validation in the way — a duplicate raises `23505` rather than
  silently producing an ambiguous badge. The resolver *also* keeps a deterministic
  lowest-`id` rule, so it stays correct against a stale cache or a client running
  ahead of the migration.
- **The link control is a pure `UPDATE` on the Calendar ladder panel**, so **no
  session insert path changed**. `LogSessionForm` was rejected as a home (it
  hard-codes `type: 'Strength'`; every ladder benchmark is an erg test) and
  `ErgLiveView` too (it could never link a historical or non-Bluetooth session, and
  `useOfflineQueue` replays queued rows verbatim, so a queued row carrying a key
  could hit the unique index on drain with no UI to report it).
- **Clear-then-set, never the reverse.** Writing the new holder while the incumbent
  still carries the key is a guaranteed `23505`. The two statements are not atomic;
  a failure between them leaves the benchmark *unlinked*, which is the safe
  direction — it can never produce a false-done, and a retry fixes it.

## What was deleted

`findMatch`, `betterFit`, the ride-elevation guard's use in the completed pass, the
grace search, and the completed-pass claim-once set are gone. The **done** matcher
collapses to a lookup.

`benchmarkKeywords`, `BENCHMARK_TERMS`, `tokenize` and `labelMatches` are **kept**
— see the rebase note above; they now serve the reschedule pass only.

`addCalendarDays` is **deliberately kept** though no production code calls it now —
it is a generic side-effect-free calendar primitive sitting beside
`diffCalendarDays`, with its tests intact. Flagging it so it reads as a decision,
not an oversight.

The two cross-entry collision tests (`P5`, `P5b`) were deleted rather than
rewritten. That is safe **only** because the scenario is now structurally
impossible: `benchmark_key` is a scalar on the row, so one session can never be a
candidate for two benchmarks.

## The tripwire, inverted

#175 deliberately shipped tests asserting the known-wrong behaviour as a tripwire
for this issue. The false-overdue one is **inverted, not deleted**:

> `AC3 leaves a benchmark overdue when a session names it but carries no link`

A `completed` session labelled `CP Test - 4min MAX (GATED)` — the real historical
label, which the old keyword resolver *did* match — dated `7/1/26`, sitting exactly
inside CP Test #1's one-day window, with `benchmark_key: null`, must still resolve
`overdue`. That test is the "no keyword fallback" decision made executable, and it
**passes after the rebase** — which is what proves pass 2 did not reintroduce a
keyword path into the done decision.

## Migration

`009_sessions_benchmark_key.sql` (+ rollback companion). Additive and reversible:
a nullable `text` column, a `COMMENT` naming `EVENT_LADDER` as the canonical slug
list, and the partial unique index. **RLS is untouched** — the four owner policies
on `sessions` are column-agnostic, so the new column is covered from day one with
zero policy edits.

**Already applied to the live database** (the spec's required order: migration
first, then Vercel — the reverse gives a PostgREST 400 on every sessions read).
Re-verified today: column and index both exist, **0 rows carry a key**. No schema
change was needed for the rebase.

## ⚠️ Expected first-deploy behaviour

**Every benchmark that currently clears by keyword will flip to OVERDUE** until it
is linked by hand. That is the correct meaning of "no link" under the no-fallback
decision, not a regression. Slightly softened by the reschedule pass: one with a
matching *future planned* session now shows `RESCHEDULED` rather than `OVERDUE`.

There is deliberately **no backfill in this PR** — it is a one-or-two-row `UPDATE`
afterwards:

```sql
update public.sessions set benchmark_key = 'cp-test-1' where id = <id>;
```

Slugs: `cp-test-1`, `cp-test-2`, `5k-tt`, `2k-test`, `tune-ups`. Verify ids against
the live table at the time rather than trusting any id quoted here.

## Deliberately left alone (follow-ups)

- **Porting the reschedule pass to `benchmark_key` too.** Would need a new migration
  — `009`'s unique index has no status predicate, so a planned row claiming a slug
  would `23505` the later completed link — plus extending `BenchmarkLinkControl` to
  offer planned sessions. Worth its own issue.
- `EVENT_LADDER.slice(0, 5)` in `CalendarView` renders indices 0–4 = **3 benchmarks
  + 2 competitions**, so `2k Test` and `1000m + tune-ups` are resolved but never
  rendered — as is the `🎯 TARGET` World Champs entry, whose colour branch is
  unreachable. Tracked in #228; panel length is a product call.
- **`.jsx` files are not linted at all** (#227). `eslint.config.js` has no config
  object whose `files` matches non-test `**/*.jsx`, so ESLint 9 skips them.
  Pre-existing and out of scope.
- No end-to-end test drives click → mutate → `onSuccess` → invalidate → re-render
  through real code (#230); each piece is covered in isolation.

## Verification (after the rebase, on `854f9f2`)

```
npm test          →  713 passed (713), 58 files
npm run lint      →  0 errors (3 pre-existing warnings)
npm run format:check → clean
npm run build     →  exit 0
coverage          →  exit 0, no threshold errors
check:design-sync →  exit 0
```

Behavioural spot-checks confirming the merge, not just the code:

- all 14 rescheduled tests pass, **including the 3 keyword-dependent ones**
  (`…future planned label does not match`, `…planned ride reschedule a distance
  benchmark`, `never reschedules a benchmark whose name yields no keywords`);
- both `compareBenchmarkSeverity` tests pass;
- `ignores a planned session even when it carries the benchmark_key` passes — pass
  1's status gate holds even against a linked planned row;
- both `CalendarView` integration tests exercising the two features **together**
  pass (`AC2 badges CP Test #2 as rescheduled and sorts it below the 5k`,
  `leaves the 5k overdue when a session names it but carries no link`).

`vite.config.js` gained two 80/80/70 per-file entries and **lowered nothing**.

Closes #188

### DIFF
--- a/supabase/migrations/009_sessions_benchmark_key.sql
+++ b/supabase/migrations/009_sessions_benchmark_key.sql
@@ -1,0 +1,34 @@
+-- 009_sessions_benchmark_key.sql
+-- Additive + reversible. Gives sessions an explicit link to a season-ladder
+-- benchmark (#188), replacing the keyword-guessing resolver: a benchmark is
+-- "done" only when a completed session carries its slug here, never because of
+-- what words happen to appear in a session label.
+--
+-- Nullable, no default, no FK (the ladder is a git-committed JS constant, not a
+-- table) and deliberately NO CHECK on the value set: a CHECK would force a
+-- migration on every ladder rename, and an unrecognised key is already inert -
+-- it matches no ladder entry, so the benchmark simply stays due.
+--
+-- RLS is untouched. The four owner policies on public.sessions
+-- (001_enable_rls.sql:29-46) are column-agnostic, so the new column is covered
+-- from day one with zero policy edits.
+
+ALTER TABLE public.sessions
+  ADD COLUMN IF NOT EXISTS benchmark_key text;
+
+COMMENT ON COLUMN public.sessions.benchmark_key IS
+  'Explicit link to a season-ladder benchmark (#188). The canonical slug list is '
+  'EVENT_LADDER in web/src/constants/schedule.js - the `key` field on the '
+  'kind=benchmark entries: cp-test-1, cp-test-2, 5k-tt, 2k-test, tune-ups. A '
+  'value outside that list is inert: it matches no ladder entry, so the '
+  'benchmark stays due and nothing breaks. At most one non-null row per user may '
+  'hold each value - enforced by sessions_one_session_per_benchmark.';
+
+-- AC4, the half that survives a Coach-over-MCP write: a duplicate link raises
+-- 23505 at the database rather than silently producing an ambiguous badge.
+-- Partial, mirroring anchors_one_live_per_key (006_context_store.sql:56-57), so
+-- every currently-unlinked row sits outside the predicate and the index cannot
+-- fail on existing data.
+CREATE UNIQUE INDEX IF NOT EXISTS sessions_one_session_per_benchmark
+  ON public.sessions (user_id, benchmark_key)
+  WHERE benchmark_key IS NOT NULL;
--- a/supabase/migrations/009_sessions_benchmark_key_rollback.sql
+++ b/supabase/migrations/009_sessions_benchmark_key_rollback.sql
@@ -1,0 +1,10 @@
+-- Rollback for 009_sessions_benchmark_key.sql. NOT auto-applied.
+--
+-- Revert the Vercel deploy FIRST. A client whose .select() names benchmark_key
+-- against a schema without the column gets a PostgREST 400 on every sessions
+-- read - which is not just a lost link, it blanks every benchmark badge.
+--
+-- Order matters: the index is built on the column, so it goes first.
+
+DROP INDEX IF EXISTS public.sessions_one_session_per_benchmark;
+ALTER TABLE public.sessions DROP COLUMN IF EXISTS benchmark_key;
--- a/web/src/components/BenchmarkLinkControl.jsx
+++ b/web/src/components/BenchmarkLinkControl.jsx
@@ -1,0 +1,116 @@
+import { useState } from 'react';
+import { COMPLETED_STATUSES } from '../constants/sessionStatus.js';
+import { THEME } from '../constants/theme.js';
+import { toISODate } from '../utils/dateFormat.js';
+
+// props:
+//   benchmarkKey    — string, required. The ladder entry's slug.
+//   benchmarkName   — string, required. For the aria-labels.
+//   linkedSessionId — number | null. Pass bench.matchedSessionId.
+//   sessions        — array of { id, date, label, status, benchmark_key }.
+//   onLink          — (sessionId | null) => void.
+//   busy            — bool. Disables the control while the write is in flight.
+//   error           — string | null. Rendered inline; keeps the select open.
+//
+// Presentational: no hooks beyond its own open/closed state, no data access.
+// The parent owns the mutation, so a failure comes back as `error` text rather
+// than being swallowed here — a 23505 means another session already claims this
+// benchmark, and the reader has to be told that.
+export default function BenchmarkLinkControl({
+  benchmarkKey,
+  benchmarkName,
+  linkedSessionId = null,
+  sessions = [],
+  onLink,
+  busy = false,
+  error = null,
+}) {
+  const [open, setOpen] = useState(false);
+  // An error keeps the select on screen, so the retry is one click away rather
+  // than behind a re-expand.
+  const expanded = open || Boolean(error);
+  const linked = linkedSessionId != null;
+
+  // The server orders sessions.date as TEXT, so the payload is not
+  // chronological. Sort here rather than trust it; the resolver no longer sorts
+  // at all. Ties on date break on the higher id — the later row of that day.
+  const candidates = sessions
+    .filter((s) => s && COMPLETED_STATUSES.includes(s.status))
+    .slice()
+    .sort((a, b) => {
+      const ai = toISODate(a.date);
+      const bi = toISODate(b.date);
+      if (ai !== bi) return ai < bi ? 1 : -1;
+      return b.id - a.id;
+    });
+
+  return (
+    <>
+      <button
+        type="button"
+        onClick={() => setOpen((v) => !v)}
+        disabled={busy}
+        aria-expanded={expanded}
+        aria-label={`Link a session to ${benchmarkName}`}
+        style={{
+          border: 'none',
+          background: 'none',
+          padding: 0,
+          cursor: busy ? 'default' : 'pointer',
+          fontFamily: 'inherit',
+          fontSize: 9,
+          letterSpacing: 1,
+          fontWeight: 700,
+          marginLeft: 6,
+          color: linked ? THEME.green : THEME.textFaint,
+        }}
+      >
+        {linked ? 'LINKED ✓' : 'LINK'}
+      </button>
+      {expanded && (
+        <select
+          aria-label={`Session for ${benchmarkName}`}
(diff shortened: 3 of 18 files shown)


skizzy1986:
**Release note — Breaking**

Benchmark completion (CP Test #1/#2, 5K TT, 2K Test, Tune-ups) no longer guesses from session labels — it now requires an explicit link between a session and its benchmark, set via a new link control on the Calendar ladder panel. Because of this, **every benchmark currently showing as done will flip to "overdue" after this deploy** until you manually relink the matching session — this is expected, not a bug, and is a one-time cleanup (5 rows). In exchange, benchmark status is now exact: a routine session can no longer accidentally clear a benchmark, and a real test logged with different wording no longer nags forever.

---

PR #232, opened 2026-08-23
docs: record the four follow-ups from #194 as drafted issues

skizzy1986:
## What changed and why

Adds `coach/followups-194.md` — the four follow-ups surfaced while working #194, written as ready-to-file issue bodies so they survive the session that found them.

Notes only. **GitHub Issues remains the tracker** per `WORKFLOW.md`; the file says so at the top and tells you to delete it once the four are filed. It is not a revival of `coach/work-orders/`.

### What's in it

| | Follow-up | Priority |
|---|---|---|
| **A** | `useErgSessions` filters `.eq('status','logged')` — matches **zero** live rows, so the hook returns nothing on every call | High |
| **B** | No CHECK constraint on `sessions.status` — nothing prevents a sixth value | Medium |
| **C** | Calendar days that are both done and cancelled render as plain "done" — 7 such dates exist | Medium |
| **D** | `useSessionLog` orders by `created_at`, not `date_iso` — insertion order, not training order | Medium |

**A is the one to look at first.** It is #194's bug inverted: another bespoke definition of "done" that drifted from the shared `COMPLETED_STATUSES`, except it counts nothing rather than too much.

A fifth item — `LogSessionForm` writing no `status` — was resolved by #221 while this sat unwritten. It is recorded as **do-not-file with the reason**, rather than deleted, so the next reader doesn't rediscover it and re-file it.

### Accuracy

Every `path:line` reference was re-checked against `main` at `9560364` immediately before committing (`App.jsx` latestErg had shifted 214 → 213), and every row count was queried from the live table rather than carried over from the #194 issue text.

Each entry carries the reasoning, the fix, acceptance criteria, and where relevant the trap:

- **C cannot be verified by looking at the calendar** — it windows today−3 → today+14 and every cancelled row is from June/July, so it needs a clocked test.
- **B and D are coupled** — if B adds `not null`, the `status == null` branch in `isCompletedStatus` stops being load-bearing and should go in the same PR.

## How to test manually

Documentation only — no code, no tests, no behaviour change.

## Checklist

- [x] CI is green — with the caveat that **Lint & Format, Test & Coverage, Build, Playwright and Lighthouse all report `skipped`, not `success`**: `Detect web changes` correctly finds no `web/` changes in a docs-only PR, so those jobs never run. Nothing regressed and nothing was bypassed, but they are not evidence of anything here. The checks that did run and pass are PR-title validation, path labelling, dependency review and the Vercel preview deploy.
- [x] Tests cover the change, or I've noted why they don't — docs-only, nothing to test
- [x] No unexpected console errors in the browser — no runtime surface changed

### DIFF
--- a/coach/followups-194.md
+++ b/coach/followups-194.md
@@ -1,0 +1,249 @@
+# Follow-ups from #194 / PR #211
+
+Four issues, ready to file. Verified against `main` at `9560364` and the live DB on
+2026-08-22. Tracking still lives in GitHub Issues per `WORKFLOW.md` — this file is
+the drafting notes, not the tracker. Delete it once the four are filed.
+
+Ranked by consequence. **A** is the only one currently producing wrong output;
+**B** and **C** are correctness-of-record; **D** is a small paired cleanup.
+
+> **Already resolved — do not file.** The original top follow-up was
+> "`LogSessionForm` writes no `status`, so hand-logged sessions are invisible to
+> CTL/ATL/TSB". **#221 fixed it** (`LogSessionForm.jsx:81` now writes
+> `status: 'completed'`), and the live table has **zero** NULL-status rows, so no
+> backfill is needed either. Superseded before it was filed.
+
+---
+
+## A. `useErgSessions` filters on `status = 'logged'`, which matches zero rows
+
+**Labels:** `bug` · **Priority:** High
+
+### Problem
+
+`web/src/hooks/useErgSessions.js:39`:
+
+```js
+.eq('status', 'logged')
+```
+
+The live `sessions` table contains **no `logged` rows at all**. Current distribution:
+
+| status | rows |
+|---|---:|
+| `actual` | 35 |
+| `cancelled` | 32 |
+| `completed` | 23 |
+| `planned` | 2 |
+| `(null)` | 0 |
+
+So this hook returns an empty set on every call, and whatever it feeds renders empty.
+
+`'logged'` is a real status value — `constants/sessionStatus.js` lists it in
+`COMPLETED_STATUSES`, and per `CLAUDE.md` it is written by the live PM5 Bluetooth
+save path. It has simply never been written yet. The bug is that this hook gates on
+that *one* value rather than the full completed set, so it silently excludes the
+58 sessions that genuinely are done.
+
+### Why it matters
+
+Same class as #194 — a bespoke definition of "done" that drifted from the shared
+one — but failing in the opposite direction. #194 counted too much; this counts
+nothing.
+
+### Fix
+
+Use the shared primitive, consistent with `useTSSHistory.js:17`:
+
+```js
+.in('status', COMPLETED_STATUSES)
+```
+
+Or `isCompletedStatus` (`utils/sessionStatus.js`, added in #211) if filtering
+client-side.
+
+### Before doing it
+
+**Establish what this hook currently feeds and whether that surface is visibly
+empty today.** If a view has been rendering blank since this landed, that is worth
+knowing — and it changes the fix from "correctness tidy-up" to "restore a broken
+screen". Confirm the before/after row count the way #211 did: write the assertion
+first, watch it fail at 0.
+
+### Acceptance criteria
+
+1. The hook returns sessions whose status is any of `COMPLETED_STATUSES`.
+2. A test asserts a `logged`, an `actual` and a `completed` row all come back, and
+   that `cancelled` and `planned` do not.
+3. The before/after count is stated in the PR, measured rather than asserted.
+
+---
+
+## B. No CHECK constraint on `sessions.status`
+
+**Labels:** `enhancement`, `database` · **Priority:** Medium
+
+### Problem
+
+`sessions.status` is plain `text`, nullable, no default, and **no CHECK constraint**
+(the only CHECK on the table is `sessions_date_parseable`). Nothing prevents a
+sixth status value — a typo from a Coach MCP write, a hand edit in the Supabase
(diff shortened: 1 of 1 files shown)


skizzy1986:
Verified B, C and D against `main` at `64bd351` and the live table. **Not filing anything** — this PR is being orchestrated, so this is evidence for that pipeline, not a competing set of issues.

**Item A is already shipped.** `useErgSessions` now uses `.in('status', COMPLETED_STATUSES)` — landed in #235 (closes #206) about an hour ago. This doc's diagnosis was independently correct, including the "matches zero live rows" claim: the table holds **0** rows with `status = 'logged'` while 23 erg sessions carry `actual`/`completed`. Worth removing from the doc before it's filed, or it'll be filed as High priority against a fixed bug.

---

## B — confirmed

```sql
select conname, pg_get_constraintdef(oid) from pg_constraint
where conrelid = 'public.sessions'::regclass and contype = 'c';
-- sessions_date_parseable | CHECK ((session_date_to_iso(date) IS NOT NULL)) NOT VALID
```

One CHECK, and it's on `date`. `status` is unconstrained nullable text, exactly as described.

Two facts that make the fix easier than the doc assumed:

| | |
|---|---|
| total rows | 92 |
| distinct statuses | 4 |
| **null status rows** | **0** |

All rows conform, so the constraint can be added and validated in one step — `not valid` isn't strictly needed. And with zero nulls, the paired `not null` is viable now rather than aspirational.

## D — confirmed

`web/src/hooks/useSessionLog.js:22` — `.order('created_at', { ascending: false })`. Insertion order, not training order, as described.

## C — the count is right, the diagnosis isn't

**7 dates carry both a completed and a cancelled session.** That number checks out exactly.

But cancelled sessions never reach `dayStatus`. `logEntriesForDate` is *deliberately* status-blind, and `utils/schedule.js:77-81` carries the contract from #194:

> CONTRACT: logEntriesForDate / dayStatus / getUpcomingSessions are STATUS-BLIND. They never inspect `status` — callers must pass an already-filtered list. Do not add a status gate here.

I checked every call site. All five pass `loggedSessions`, which `useSessionLog.js:87` builds via `if (isCompletedStatus(e.status))`:

```
CalendarView.jsx:70      dayStatus(d, today0, loggedSessions)
OverviewView.jsx:78      getUpcomingSessions(nowTick, loggedSessions)
OverviewView.jsx:139     logEntriesForDate(...)
schedule.js:95, :151     internal, receive the caller's list
```

So on those 7 dates the day renders `done` **because a session genuinely was completed** — the cancelled sibling is correctly irrelevant. And the 16 dates carrying *only* cancelled sessions render `missed`, which is also correct.

What's actually left is a **UX gap, not a correctness defect**: the cancelled session is invisible, so the calendar gives no sign you planned something and dropped it. Real, but an enhancement — filing it at Medium as a done/cancelled correctness bug would send someone hunting for a defect #194 and #211 already closed.

## Not in the doc: one comment is already stale

`web/src/utils/sessionStatus.js:4-9` states:

> null/undefined is admitted, and that branch is **LOAD-BEARING** — not defensive padding. `LogSessionForm.jsx:72-81` inserts a session without naming a `status` field at all…

That stopped being true. #221 made `LogSessionForm` write `status: 'completed'` explicitly (now line 81), and the table holds **0** null-status rows. The doc anticipates this going stale *if* B adds `not null` — it is stale **now**, independently of B, and the comment actively misdescribes current behaviour.

Worth folding into B rather than tracking separately, since anyone touching that constraint will read that comment first.

---

skizzy1986:
Closing unmerged — all four follow-ups are now accounted for, which is the condition this file set for its own deletion ("Delete it once the four are filed").

| | Follow-up | Outcome |
|---|---|---|
| **A** | `useErgSessions` filters `status='logged'`, matching zero rows | Fixed by #235 (closes #206) — landed before this was actioned |
| **B** | No CHECK constraint on `sessions.status` | Fixed by #243 — migration `010`, `NOT NULL` + `sessions_status_known` |
| **C** | Mixed cancelled+done days leave no trace on the calendar | Filed as #242 |
| **D** | `useSessionLog` orders by `created_at` | Fixed by #243 |

Closing rather than merging because the notes never landed on `main` — `coach/followups-194.md` exists only on `claude/orchestrate-194-yonfxh`. Merging first to delete second would put a file into history purely to remove it. Closing achieves what the file asked for, and the branch is deleted alongside.

The content isn't lost: each follow-up's reasoning now lives where it's actionable — B and D in #243's body and in the migration's header comment, C in #242 with its constraints and the clocked-test warning intact.

### Two corrections the build surfaced

**A's diagnosis was right but its priority had expired.** Marked High here; #235 had already shipped `.in('status', COMPLETED_STATUSES)` by the time this was picked up. Verified rather than assumed — the live table still holds zero `logged` rows.

**C's diagnosis was wrong, though its row count was exact.** The notes described it as done-vs-cancelled rendering incorrectly. It isn't: cancelled sessions never reach `dayStatus` at all. All five call sites of the status-blind helpers pass `loggedSessions`, and `logDisplaySessions` — the only list carrying cancelled rows — reaches `LogView` and nothing else. Those 7 days render `done` because a session genuinely was completed, and the 16 cancelled-only dates render `missed`. Both correct. #242 is filed as an enhancement so nobody hunts for a defect #194 and #211 already closed.

One thing the notes didn't catch, worth recording since it's the same class of drift they were tracking: the "LOAD-BEARING" comment on `utils/sessionStatus.js`'s null branch was **already false** before any of this — #221 made `LogSessionForm` write `status: 'completed'` explicitly. The notes predicted it would go stale *if* B added `not null`; it had gone stale independently. Fixed in #243.

---

PR #235, opened 2026-08-23
fix: show the erg history ErgView was filtering away (#206)

### LINKED ISSUE #206: useErgSessions filters status='logged', which no session row matches — ErgView's list is empty

skizzy1986:
Closes #206

## What changed and why

`useErgSessions` queried `.eq('status', 'logged')`. Every other consumer of *"did this session actually happen?"* reads `COMPLETED_STATUSES` — `useTSSHistory`, `useCoach`, `MobileAnalytics`, `benchmarkStatus`, and the shared `isCompleted()` helper. This hook was the **only** place that hardcoded a single status, and it picked the one the data doesn't use.

Counted against the live table just now:

| status | rows | of which erg |
|---|---:|---:|
| `actual` | 35 | 5 |
| `cancelled` | 32 | 11 |
| `completed` | 23 | 18 |
| `planned` | 2 | 0 |
| **`logged`** | **0** | **0** |

So the filter matched **nothing**. ErgView has been rendering an empty list while **23 completed erg sessions** sat in the table.

`'logged'` isn't dead code — `ErgLiveView.jsx:618` writes it on the PM5 Bluetooth save path. It simply has no rows yet, which is exactly why this survived review: the code path that *produces* the status exists, so the filter reads plausible. Only counting rows shows it never matches.

Swapped to `.in('status', COMPLETED_STATUSES)`. Nothing else in the hook changes — the `date_iso` ordering from #187 and the 50-row limit are untouched.

## Tests

Three regression tests. The mock chain gains `.in()` / `.eq()` capture so the **filter itself** is assertable, rather than inferred from whatever rows the mock returns:

- `filters on every completed status, not just one` — asserts `.in('status', COMPLETED_STATUSES)`
- `never narrows the query to a single status` — asserts `.eq` is never called with `status`, so the old shape can't come back in a different spelling
- `returns bulk-imported history that carries actual/completed` — an `actual`, a `completed` and a `logged` row all survive to the hook's output

**Verified both ways:** restoring `.eq('status','logged')` fails two of the three; the fix passes all 22 in the file.

## How to test manually

Open the Erg tab. Before this change it lists nothing; after, it lists the erg history (23 sessions in the live data).

## Checklist

Verified locally: `lint` **0 errors**, `format:check` clean, **716 tests across 58 files**, `build` exit 0, `coverage` exit 0, `check:design-sync` exit 0.

---

### DIFF
--- a/web/src/hooks/__tests__/useErgSessions.test.js
+++ b/web/src/hooks/__tests__/useErgSessions.test.js
@@ -6,6 +6,8 @@ import React from 'react';
 const fromMock = vi.fn();
 const orderMock = vi.fn();
 const limitMock = vi.fn();
+const eqMock = vi.fn();
+const inMock = vi.fn();
 
 vi.mock('../../supabaseClient.js', () => ({
   supabase: {
@@ -15,6 +17,7 @@ vi.mock('../../supabaseClient.js', () => ({
 
 import { useErgSessions } from '../useErgSessions.js';
 import { wattsToPace500 } from '../../utils/pace.js';
+import { COMPLETED_STATUSES } from '../../constants/sessionStatus.js';
 
 function makeWrapper() {
   const client = new QueryClient({
@@ -52,7 +55,14 @@ function mockQuery(data, error = null, cp = 205) {
     }
     const chain = {
       select: () => chain,
-      eq: () => chain,
+      eq: (...args) => {
+        eqMock(...args);
+        return chain;
+      },
+      in: (...args) => {
+        inMock(...args);
+        return chain;
+      },
       order: (...args) => {
         orderMock(...args);
         return chain;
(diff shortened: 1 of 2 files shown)


skizzy1986:
**Release note draft — Fixed**

The Erg tab was showing no history at all — a leftover status filter meant it only ever matched sessions logged live over PM5 Bluetooth, so the 23+ completed erg sessions already in your training log never appeared. The Erg tab now shows your full completed erg history again, matching the same "completed" logic used everywhere else in the dashboard.

---

PR #237, opened 2026-08-23
chore: clear the dead code the .jsx lint gate exposed

skizzy1986:
Follow-up to #234 (#227).

## What changed and why

#234 made ESLint visit 35 non-test `.jsx` files for the first time and deliberately left **12 `no-unused-vars` warnings** standing, so that PR stayed a lint-config change rather than a dead-code sweep across the entry shell. This is the sweep.

`npm run lint` now reports **nothing at all** — 0 problems, down from 12 warnings.

## Nine were genuinely dead

Each verified to have **exactly one reference** in its file — its own declaration — before removal:

| File | Symbol | Why it's dead |
|---|---|---|
| `App.jsx` | `lazy`, `Suspense` | imported from react, never used |
| `App.jsx` | `ErgTooltip`, `C` | imported, never used |
| `App.jsx` | `progTab` / `setProgTab` | `useState` never read — `ProgramView` has owned its own sub-tab state since the decomposition |
| `App.jsx` | `isMid` | computed, never used (`isWide` still is) |
| `MobileApp.jsx` | `PlaceholderCard` | local component, not exported, unreferenced |
| `LoadTooltip.jsx` | `label` | unused Recharts tooltip prop |

## Three were intent, not accident — so the code stayed and the rule moved

**`strengthDraft.js` — two `catch (e)`.** The binding is unused and the comment above each already explains the swallow, so these become the optional-catch-binding form `catch {`. No behaviour change.

**`useOfflineQueue.js` — `const { _queuedAt, ...rest } = session`.** This is the deliberate *omit-a-key* idiom: a rest-spread filter that strips a local-only field before the insert. **Deleting it would reintroduce `_queuedAt` into the row sent to Supabase.** `no-unused-vars` has `ignoreRestSiblings` for exactly this case, so that's now on.

`varsIgnorePattern: '^_'` is set alongside it, so the underscore convention means the same thing for variables as `argsIgnorePattern` already made it mean for arguments — the author had already signalled intent with the `_` prefix; the config just didn't honour it.

## Risk

**Nothing here changes runtime behaviour.** Every removal is an unreferenced binding, and the two rule options only affect what gets *reported*. The full diff is 16 insertions / 27 deletions across 5 files, with no incidental reformatting.

## Checklist

- [x] Tests cover the change, or I've noted why they don't — *deletions of unreferenced bindings; the existing 716 tests are the regression net*

Verified locally: `lint` **0 problems**, `format:check` clean, **716 tests across 58 files**, `build` exit 0, `coverage` exit 0, `check:design-sync` exit 0.

---

### DIFF
--- a/web/eslint.config.js
+++ b/web/eslint.config.js
@@ -49,7 +49,18 @@ export default [
     rules: {
       ...reactPlugin.configs.recommended.rules,
       ...reactHooksPlugin.configs.recommended.rules,
-      'no-unused-vars': ['warn', { argsIgnorePattern: '^_' }],
+      // ignoreRestSiblings covers the deliberate omit-a-key idiom
+      // (`const { _queuedAt, ...rest } = session`), which is a rest-spread
+      // filter, not dead code. varsIgnorePattern extends the existing `_`
+      // convention from arguments to variables so the two agree.
+      'no-unused-vars': [
+        'warn',
+        {
+          argsIgnorePattern: '^_',
+          varsIgnorePattern: '^_',
+          ignoreRestSiblings: true,
+        },
+      ],
       'no-console': 'off',
       'react/prop-types': 'off',
       'react/react-in-jsx-scope': 'off',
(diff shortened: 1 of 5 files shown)


skizzy1986:
Internal change — no release note. This is a dead-code/lint-config cleanup (unused variables removed, `ignoreRestSiblings`/`varsIgnorePattern` added) with no runtime behavior change.

---

PR #238, opened 2026-08-23
feat: render the whole event ladder, not the first five (#228, #215)

### LINKED ISSUE #215: The calendar ladder picks its visible five by position, so overdue benchmarks past index 4 stay invisible (bites Feb 2027)
Found during validation of #192. **Pre-existing** — `EVENT_LADDER.slice(0, 5)` predates that work; #192 only makes the consequence legible by sorting within the slice.

## Problem

`web/src/views/CalendarView.jsx` selects the visible five benchmarks by **ladder position**, then sorts those five by severity. Slicing before sorting is deliberate (#192 pins it with a test) because sorting first would change *which* events are visible — a product decision that was explicitly out of scope there.

The consequence is that a benchmark outside the first five can go overdue and never render.

## Dated evidence

`2k Test` is `EVENT_LADDER` index **5**; `1000m + 1-min tune-ups` is index **6**. Both sit outside the visible window. Running the real resolver at `today=2027-02-15`:

```
idx0 overdue  CP Test #1               idx5 overdue  2k Test                 <- invisible
idx1 overdue  CP Test #2               idx6 overdue  1000m + 1-min tune-ups  <- invisible
idx2 overdue  5k Time Trial
VISIBLE: 0:overdue | 1:overdue | 2:overdue | 3:quiet | 4:quiet
```

Indices 0–2 never expire, so by Feb 2027 the panel shows **three permanently-stale 2026 warnings while both live pre-peak benchmarks are hidden**. With a planned reschedule, `2k Test` correctly resolves to `scheduled` — and is still never rendered.

That is precisely the failure #192's own AC5 test comment names: *"five permanent warnings in ladder order is the same wallpaper the feature exists to remove."*

## Why this needs a decision, not a patch

The obvious fix — select the visible five by severity across the whole ladder — is what #192's spec deliberately forbade, because it changes which events the calendar surfaces. Options worth weighing:

1. Select by severity across the full ladder (loudest five, whatever their position).
2. Keep positional selection but let overdue/scheduled entries beyond index 4 append below the five.
3. Widen the slice.
4. Expire stale benchmarks so 2026 entries stop occupying the panel indefinitely — arguably the real root cause, since indices 0–2 hold their slots forever.

## Maintenance note

The `toHaveLength(1)` / `toHaveLength(2)` badge-count assertions in `web/src/views/__tests__/CalendarView.benchmark.test.jsx` are only time-stable **because** the slice currently hides indices 5+. Whoever widens the slice inherits those and will need to revisit them.

## Timing

No urgency until the Jan 2027 windows approach, but worth deciding before then rather than discovering it live.

### LINKED ISSUE #228: feat: 2k Test and 1000m tune-ups are resolved but never rendered, so they can only be linked by Coach
Follow-up to #188 (PR #226). Surfaced by the Reality Checker at that feature's
Gate 3, and explicitly out of scope there.

## The problem

`web/src/views/CalendarView.jsx` renders `EVENT_LADDER.slice(0, 5)` — indices 0–4.
That slice is **3 benchmarks + 2 competitions**:

| Index | Entry | `kind` | Rendered? |
|---|---|---|---|
| 0 | CP Test #1 (4-min) | benchmark | yes |
| 1 | CP Test #2 (2nd duration) | benchmark | yes |
| 2 | 5k Time Trial | benchmark | yes |
| 3 | Erg Power Series | competition | yes |
| 4 | C2 Monthly | competition | yes |
| 5 | **2k Test** | **benchmark** | **no** |
| 6 | **1000m + 1-min tune-ups** | **benchmark** | **no** |

So of the 5 `kind: 'benchmark'` entries, only **3** are badge-visible. The hook
resolves all 5 correctly — `useBenchmarkStatuses` runs over the whole ladder — but
indices 5 and 6 are sliced off before render.

## Why it matters now

Before #188 this was only a missing badge. After #188 it is also a missing
**affordance**: the new `BenchmarkLinkControl` renders inside the same slice, so
`2k-test` and `tune-ups` have no in-app way to be linked to a session at all. They
can only ever be linked by Coach writing `sessions.benchmark_key` directly over
MCP.

Both are real season benchmarks (`~Mid Jan 27` and `Late Jan 27`), so this becomes
a live problem around January rather than a theoretical one.

## What to weigh at spec time

This is a product call about panel length, not a bug fix — the Calendar ladder
panel is deliberately short. Options worth costing:

1. Widen the slice (show more ladder entries on the Calendar).
2. Keep the panel short but surface the unrendered benchmarks somewhere else —
   `ProgramYear.jsx` already renders the **full** ladder as static content and
   would be a natural host for link controls.
3. Show only `kind: 'benchmark'` entries in the panel, dropping the two
   competitions, so all 5 benchmarks fit in the same visual budget.

## Gotcha for whoever implements

`borderBottom: i < 4` in `CalendarView.jsx` is **hard-coded to the slice length** —
it draws a separator on every row but the last. Widening the slice without
changing that constant leaves a stray divider under the wrong row.

skizzy1986:
Closes #228
Closes #215

## What changed and why

The Calendar ladder panel capped itself at `EVENT_LADDER.slice(0, 5)`. Indices 0–4 are three benchmarks and two competitions, so **four of the nine entries never rendered at all**:

| idx | kind | name | date |
|---:|---|---|---|
| 5 | `benchmark` | 2k Test | ~Mid Jan 27 |
| 6 | `benchmark` | 1000m + 1-min tune-ups | Late Jan 27 |
| 7 | **`TARGET`** | World Rowing Virtual Indoor Champs | Late Feb 27 |
| 8 | `optional` | WR Virtual Indoor Sprints | Early Mar 27 |

Three consequences:

- **The `TARGET` entry — the season's whole point — was invisible**, and the distinct `#ff2d55` colour branch this view carries for `kind === 'TARGET'` was unreachable code.
- **The cut was positional, so severity couldn't rescue it.** A benchmark going overdue deep in the ladder could never surface however loud it got — the 2k Test would stay hidden through the entirety of its overdue window (#215, "bites Feb 2027").
- **Only 3 of the 5 benchmarks were badge-visible**, so the other two could only be linked by Coach writing `benchmark_key` directly — there was no row to click. All five now carry a link control.

## Why the cap existed, and why nine rows is now fine

The cap was a deliberate decision, not an oversight — the test comment said it plainly: *"five permanent warnings in ladder order is the same wallpaper the feature exists to remove."*

What makes nine rows survivable is that **severity ordering already landed in #192/#214**: overdue → upcoming → scheduled → quiet, ties broken on ladder index. The panel leads with whatever is actionable, and the quiet remainder sits below in ladder order. The cap was solving a problem the sort now solves better — and solving it in a way that *created* #215.

`borderBottom` moves from a hardcoded `i < 4` to `i < visibleEvents.length - 1`, which is exactly what `ProgramYear.jsx:273` already does for the same ladder. **The two views that render `EVENT_LADDER` are consistent now** — CalendarView was the odd one out.

## Tests: two acceptance criteria inverted, not deleted

**`AC5 sorts within the visible five without changing which five they are`** existed specifically to assert a deep-overdue entry must *not* surface. That's now the opposite of desired behaviour, so it's inverted and renamed — `AC5 surfaces a benchmark that goes overdue deep in the ladder (#215)` — asserting it appears *and* takes row 0.

Two others were made derive-from-the-ladder rather than hardcoded, so adding a benchmark extends them instead of silently passing:

- the severity test computes its expected quiet remainder from `EVENT_LADDER` instead of asserting three rows;
- the link-control test builds its expected aria-labels from every `kind === 'benchmark'` entry.

Added one test that the `TARGET` entry renders at all.

## Scope note

No mobile change needed — I checked, and the ladder panel exists only in the desktop `CalendarView`. The `slice(0, 5)` in `MobileAnalytics.jsx` is an unrelated "5 most recent sessions" list.

The panel is now roughly 80% taller. That's the accepted trade for the entries being reachable; worth an eye on the Vercel preview.

## Checklist

Verified locally: `lint` **0 problems**, `format:check` clean, **717 tests across 58 files**, `build` exit 0, `coverage` exit 0, `check:design-sync` exit 0.

---

### DIFF
--- a/web/src/views/CalendarView.jsx
+++ b/web/src/views/CalendarView.jsx
@@ -81,11 +81,14 @@ export default function CalendarView({ loggedSessions, isWide }) {
   }
   const todayMode = firstMode;
   const todayCycle = MICROCYCLE[todayMode] || MICROCYCLE.home;
-  // Slice BEFORE sort: which five entries are visible stays a ladder decision,
-  // only their order is a severity one. .slice() copies, so the memoized hook
-  // result is never mutated.
+  // The whole ladder renders, ordered by severity (#228). It used to be capped
+  // at the first five by position, which hid the 2k Test, the 1000m tune-ups
+  // and the TARGET champs entirely — and meant a benchmark going overdue deep
+  // in the ladder could never surface, however loud (#215). Severity order is
+  // what keeps nine rows from being wallpaper: anything actionable floats.
+  // .slice() still copies, so .sort() never mutates the memoized hook result.
   const visibleEvents = benchmarkStatuses
-    .slice(0, 5)
+    .slice()
     .sort(compareBenchmarkSeverity);
   return (
     <>
@@ -256,7 +259,8 @@ export default function CalendarView({ loggedSessions, isWide }) {
                 gap: 10,
                 marginBottom: 6,
                 paddingBottom: 6,
-                borderBottom: i < 4 ? '1px solid #3e3e5a' : 'none',
+                borderBottom:
+                  i < visibleEvents.length - 1 ? '1px solid #3e3e5a' : 'none',
               }}
             >
               <div
--- a/web/src/views/__tests__/CalendarView.test.jsx
+++ b/web/src/views/__tests__/CalendarView.test.jsx
@@ -142,9 +142,10 @@ describe('CalendarView', () => {
     );
   });
 
-  // AC5 — five permanent warnings in ladder order is the same wallpaper the
-  // feature exists to remove. What is actionable has to be at the top.
-  it('AC5 sorts the visible rows overdue → upcoming → the rest', () => {
+  // AC5 — nine permanent warnings in ladder order is the same wallpaper the
+  // feature exists to remove. What is actionable has to be at the top; severity
+  // order is what makes rendering the whole ladder (#228) survivable.
+  it('AC5 sorts every row overdue → upcoming → the rest', () => {
     const states = quietStates();
     states[4] = { ...states[4], status: 'overdue', daysOverdue: 12 };
     states[1] = { ...states[1], status: 'upcoming', daysUntilStart: 3 };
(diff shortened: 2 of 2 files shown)


skizzy1986:
**Release note (Fixed)**

The Calendar event ladder was only ever showing its first 5 entries — which meant the World Rowing Virtual Indoor Champs (your actual season target) never appeared at all, along with the 2k Test, the 1000m tune-ups, and the WR Virtual Indoor Sprints. An overdue benchmark buried further down the list also couldn't surface no matter how overdue it got. The ladder now renders all 9 events, sorted so whatever's most urgent (overdue, then upcoming) floats to the top — the panel is correspondingly taller. All 5 benchmarks now also have a linkable control, not just the first 3.


---

PR #231, opened 2026-08-23
fix: install the web workspace in the SessionStart hook, not the repo root

skizzy1986:
## What changed and why

`.claude/hooks/session-start.sh` cd'd to `$CLAUDE_PROJECT_DIR` and ran a bare `npm install`. The repo root `package.json` declares only `husky` — the app is the `web/` workspace child — so every remote session began with `web/node_modules` absent and `vitest`, `eslint` and `vite` unavailable. `npm test` returned `sh: 1: vitest: not found`.

The hook's closing message advertised `lint: npm run lint | test: npm test | build: npm run build`, none of which could run from where it left you. A session could not execute any of the three gates CLAUDE.md requires before pushing.

Now `npm install --prefix web`, with the closing message corrected to say the commands run from `web/`.

**CI was never affected** — `ci-web.yml` sets `defaults.run.working-directory: web`. This is a dev-environment-only fix; no application code changes.

## How to test manually

From a fresh checkout:

```bash
npm install --prefix web     # populates web/node_modules (475 packages)
cd web && npm run lint && npm run format:check && npm test && npm run build
```

Before this change the install placed only `husky` at the root and every command after it failed.

## Checklist

- [x] CI is green (lint, test, build)
- [x] Tests cover the change, or I've noted why they don't — *shell hook, not covered by Vitest; verified by running all four gates end to end*
- [x] `npm run build` passes locally
- [x] No unexpected console errors in the browser — *no application code changed*

Verified locally after the fix: lint 0 errors (3 pre-existing `no-unused-vars` warnings), `format:check` clean, **689 tests passed across 56 files**, `npm run build` exit 0.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q8RCQ5d8TKHa8DdzMU1Nmc)_

PR #241, opened 2026-08-23
fix: use npm ci in the SessionStart hook so it stops rewriting the lockfile

skizzy1986:
Follow-up to #231.

## What changed and why

#231 fixed the hook to install the `web/` workspace instead of the repo root, but kept `npm install`. On a runner whose npm differs from the one that generated `package-lock.json`, **`install` rewrites the lockfile.** Here it stripped the `"libc"` metadata off every optional platform dependency:

```
10  -      "libc": [
 6  -        "glibc"
 4  -        "musl"
```

No `version`, `resolved` or `integrity` change — purely a silent downgrade of the platform hints npm uses to skip incompatible optional deps. It surfaced as uncommitted changes in a tree nobody had touched, repeatedly across today's session, and committing it would have quietly degraded the lockfile for every other machine.

`npm ci` installs strictly from the lockfile and never writes to it.

## Verified, not assumed

**Claim 1 — it installs the deps and leaves the lockfile alone:**

```
rm -rf web/node_modules && npm ci --prefix web

lockfile sha before: a62b7bdaa631a524
lockfile sha after:  a62b7bdaa631a524   ← unchanged
git sees 0 changes to the lockfile
475 packages installed; vitest / eslint / vite all present (3/3)
```

**Claim 2 — the trade-off is that it aborts on a lockfile mismatch.** Tested by adding a dep to `package.json` without updating the lock:

```
npm error code EUSAGE
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json ... are in sync.
npm error Missing: left-pad@1.3.0 from lock file
exit 1
```

Under `set -euo pipefail` that aborts the hook. **That's the right outcome** — a session built on a lockfile nobody committed is worse than a session that tells you to fix the lockfile. Both behaviours are written into the comment so the next reader doesn't have to rediscover why it's `ci`.

## A note on the CI signal

This changes `.claude/` only, so `Detect web changes` correctly finds no `web/` changes and **Lint & Format, Test & Coverage, Build, Playwright and Lighthouse all report `skipped`, not `success`** — the same situation as #231. Branch protection counts `skipped` as passing (that's the documented reason `ci-web.yml` filters at the job level rather than with `on.paths`), but those checks are not evidence of anything here.

What actually stands behind this change: the two verification runs above, `bash -n` on the hook, and a full local gate run on an untouched `web/` — lint 0 problems, 717 tests across 58 files, build exit 0, design-sync guard exit 0.

## How to test manually

```bash
sha256sum web/package-lock.json
rm -rf web/node_modules
npm ci --prefix web
sha256sum web/package-lock.json   # identical
git status --short                # clean
```

With `npm install` the same sequence leaves 30 deletions in the lockfile.

## Checklist

- [x] CI is green (lint, test, build) — *see the note above: skipped, not run, and correctly so*
- [x] Tests cover the change, or I've noted why they don't — *shell hook; verified by running the command both ways and diffing the lockfile*

---

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -14,6 +14,17 @@ echo "[session-start] Installing npm dependencies..."
 # The app is the web/ workspace child. The repo root package.json declares
 # only husky, so a bare `npm install` here leaves web/node_modules absent and
 # vitest/eslint/vite unavailable for the whole session.
-npm install --prefix web
+#
+# `ci`, not `install`: install REWRITES package-lock.json whenever the runner's
+# npm differs from the one that generated it — here it stripped every "libc"
+# field off the optional platform deps, which is a silent lockfile downgrade
+# that then shows up as uncommitted changes in an otherwise untouched tree.
+# `ci` installs strictly from the lockfile and never writes to it.
+#
+# It also fails loudly if package.json and the lockfile disagree, rather than
+# quietly resolving something new. Under `set -e` that aborts the hook, which
+# is the right outcome: a session built on a lockfile nobody committed is worse
+# than a session that tells you to fix the lockfile.
+npm ci --prefix web
 
 echo "[session-start] Done. Run from web/: npm run lint | npm test | npm run build"


skizzy1986:
Internal change — no release note. This only touches the `.claude/` SessionStart hook (swaps `npm install` for `npm ci` so the dev-session bootstrap stops rewriting the lockfile); no application code changed.

---

Issue #242, opened 2026-08-23
feat: show on the calendar that a day had a cancelled session, not just a completed one


skizzy1986:
Follow-up **C** from `coach/followups-194.md` (PR #232). Filed as an enhancement rather than a bug — see *Not a correctness defect* below, which corrects the framing in the drafting notes.

## Problem

Seven dates carry both a completed session and a cancelled one. The calendar renders them as an ordinary done day, so the fact that something was prescribed and dropped leaves no trace.

| date | cancelled | done |
|---|---:|---:|
| 6/23/26 | 1 | 1 |
| 6/25/26 | 1 | 2 |
| 6/27/26 | 1 | 1 |
| 7/3/26 | 2 | 1 |
| 7/13/26 | 1 | 1 |
| 7/14/26 | 1 | 1 |
| 7/20/26 | 2 | 1 |

That is exactly the scenario that set #211's direction: a prescription skipped, something else done instead. The Log records both; the calendar shows only the half that happened.

## Not a correctness defect

The notes framed this as done-vs-cancelled rendering wrongly. Verified against `main` — it isn't. Cancelled sessions never reach `dayStatus` at all.

All five call sites of the status-blind helpers pass `loggedSessions`, which `useSessionLog.js:87` builds through `isCompletedStatus`:

```
CalendarView.jsx:70      dayStatus(d, today0, loggedSessions)
OverviewView.jsx:78      getUpcomingSessions(nowTick, loggedSessions)
OverviewView.jsx:139     logEntriesForDate(...)
schedule.js:95, :151     internal, receive the caller's list
```

`logDisplaySessions` — the only list carrying cancelled rows — reaches `LogView` and nothing else.

So those 7 days render `done` because a session genuinely was completed, and the 16 dates carrying *only* cancelled sessions render `missed`. Both correct. What's missing is purely additive: surfacing the dropped session.

## Constraints

1. **`utils/schedule.js` must stay status-blind.** The contract at `schedule.js:77-81` is pinned by a characterization test and exists because a second status gate in a second layer is what caused #194. Do not teach the date helpers about status — pass cancelled data as a **separate prop**.
2. **The day is genuinely two states at once.** How one cell renders both is a design question — combined badge, count suffix, precedence — with knock-ons through `statusColor` / `statusLabel` and the `logged.length` count in `CalendarView.jsx`.

## Cannot be verified by looking at the calendar

[...]

PR #243, opened 2026-08-23
fix: constrain sessions.status and order the Log by training date (#232 B+D)

skizzy1986:
Refs #232 — that is the drafting-notes PR these follow-ups came from, not an issue, so it does not auto-close. Follow-up C is filed separately as #242.

## What changed and why

Builds follow-ups **B** and **D** from `coach/followups-194.md`: `sessions.status` can no longer hold an unknown or missing value, and the Log orders by when training happened rather than when the row was inserted.

They ship together because they are coupled through `utils/sessionStatus.js` — B removes the last way a null can reach `isCompletedStatus`, so the branch admitting one has to go in the same change.

### Status of all four follow-ups

| | Follow-up | Outcome |
|---|---|---|
| **A** | `useErgSessions` filters `status='logged'`, matching zero rows | Already fixed by #235 before this ran — verified, not re-filed |
| **B** | No CHECK constraint on `sessions.status` | **This PR** |
| **C** | Mixed cancelled+done days leave no trace on the calendar | Filed as #242 (enhancement, not a bug — see below) |
| **D** | `useSessionLog` orders by `created_at` | **This PR** |

### B — `sessions.status` was unconstrained

Plain nullable text, no CHECK. A Coach MCP typo, a hand edit in the Supabase dashboard, or a future write path inventing `'skipped'` were all accepted silently. Consumers already fail closed on an unknown value, but that is a mitigation — the hazard was that the column accepted it.

Migration `010_sessions_status_constraint.sql` adds `NOT NULL` plus a `sessions_status_known` allow-list, with a rollback file alongside.

**Deliberately no DEFAULT.** The drafting notes floated one. A default is the exact failure the constraint exists to close: a writer that forgets `status` would land silently as phantom training (`'completed'`) or a phantom prescription (`'planned'`) instead of raising. All three insert paths already state their intent.

**Added VALID in one step**, not `NOT VALID` + `VALIDATE` — all 92 rows conform and the scan is trivial, so the two-step form would only leave a second step to forget.

### D — `useSessionLog` ordered by insertion, not training date

`.order('created_at', …)` meant a session imported or edited late sorted as "most recent" however long ago it was actually done. Now `date_iso` desc with `NULLS LAST` plus an `id` tiebreak, matching `useErgSessions.js:55-56`.

## Before / after — measured, not asserted

Queried against the live table, completed statuses only, `type <> 'Test'`:

| pos | before (`created_at`) | after (`date_iso`) |
|---|---|---|
| **1** | **`7/14/26` — UT1 Steady 50min @ 158W** | **`8/6/26` — UT1 Steady 50min @ 164W** |
| 2 | `8/4/26` — 40min progressive build | `8/4/26` — 40min progressive build |
| 3 | `8/2/26` — UT1 45min Ramp SR | `8/2/26` — UT1 45min Ramp SR |
| 4 | `7/30/26` — UT1 Steady 45min @ 157W | `7/30/26` — UT1 Steady 45min @ 157W |

The latest-erg tile was naming a session **23 days staler** than the actual most recent one, with three newer sessions sitting below it. Positions 2–4 are unchanged, so this is the only visible movement in the recent-4 list. "Last sRPE" is unaffected — every row in the top 4 has `srpe` null under both orderings.

## Migration verification

Applied via `apply_migration`, then read back:

```
sessions_status_known => CHECK (status = ANY (ARRAY['actual','completed','logged','planned','cancelled']))
status_nullable       => NO
total_rows            => 92
distribution          => actual=35, cancelled=32, completed=23, planned=2
```

Constraint is **valid**, not `NOT VALID`. Row count and distribution unchanged.

Probed all three rejection paths against the live table — unknown status, explicit NULL, and omitted status. All three rejected, **92 rows and 0 probe rows leaked** afterwards.

Supabase security + performance advisors run post-DDL: only pre-existing findings, none touching `sessions`, nothing new introduced.

## Not in the original notes

Two things this PR fixes that the drafting notes didn't record:

1. **`utils/sessionStatus.js`'s comment was already wrong**, independent of the migration. It declared the null branch "LOAD-BEARING" and cited `LogSessionForm.jsx:72-81` inserting without a status — #221 made that line write `status: 'completed'` explicitly. The branch was dead before the constraint removed the last way to reach it.
2. **Two supabase mocks would have thrown.** `useSessionLog.test.js` and `App.test.jsx` both defined `order` as a terminal Promise, so the second chained `.order()` called `.order` on a Promise. Both are now chainable, and `useSessionLog.test.js` records the arguments so the ordering contract is asserted directly rather than inferred from row order.

## C was re-characterised

The notes framed C as a done-vs-cancelled rendering defect. Verified — it isn't: cancelled sessions never reach `dayStatus`, because all five call sites of the status-blind helpers pass `loggedSessions`, and `logDisplaySessions` reaches only `LogView`. Those 7 days render `done` correctly and the 16 cancelled-only dates render `missed` correctly. What's missing is that a dropped session leaves no trace — additive UI work. Filed at #242 with that framing so nobody hunts for a defect #194 and #211 already closed.

## How to test manually

Open Overview and the Log on the preview. The latest-erg tile should name the **8/6/26** session, not 7/14/26, and the Log should read in training-date order.

## Checklist

- [x] Tests cover the change, or I've noted why they don't — 720 tests pass across 58 files; new `#232-D` block asserts the ordering contract, the late-created/early-dated case, and null exclusion; `sessionStatus.test.js` AC3 inverted to pin the removed behaviour. Coverage: lines 84.76% / functions 79.87% / branches 77.67%, all above the ratchet (80 / 75 / 72)
- [x] No unexpected console errors in the browser — no new runtime surface; the ordering change is a query argument

---
_Generated by [Claude Code](https://claude.ai/code/session_017qkwjR61snwGxwSgwLKf3W)_

### DIFF
--- a/supabase/migrations/010_sessions_status_constraint.sql
+++ b/supabase/migrations/010_sessions_status_constraint.sql
@@ -1,0 +1,49 @@
+-- 010_sessions_status_constraint.sql
+-- Additive + reversible. Closes the gap that let sessions.status hold anything at
+-- all: the column was plain nullable text with no CHECK, so a Coach MCP typo, a
+-- hand edit in the Supabase dashboard, or a future write path inventing 'skipped'
+-- would all be accepted silently.
+--
+-- Consumers already fail closed on an unknown value (utils/sessionStatus.js is an
+-- allow-list, useTSSHistory filters server-side with .in(...)), so a bad value
+-- makes a session go visibly missing rather than counting as phantom training.
+-- That is a mitigation, not a fix - the hazard is that the column accepts it.
+--
+-- Safe on current data, verified before writing rather than assumed: 92 rows,
+-- four distinct statuses (actual 35, cancelled 32, completed 23, planned 2), and
+-- ZERO nulls. Every row already satisfies both constraints below.
+--
+-- No writer breaks either. All three insert paths already name a status:
+--   web/src/components/LogSessionForm.jsx:81  -> 'completed' (since #221)
+--   web/src/views/ErgLiveView.jsx             -> 'logged'
+--   web/src/hooks/useOfflineQueue.js:75       -> spreads the ErgLiveView row
+--
+-- RLS is untouched. The four owner policies on public.sessions
+-- (001_enable_rls.sql:29-46) are column-agnostic, so nothing here needs a policy
+-- edit.
+
+-- Deliberately NO DEFAULT. A default is the exact failure this migration exists
+-- to close: a write path that forgets `status` would land silently as phantom
+-- training ('completed') or a phantom prescription ('planned') instead of raising.
+-- Every current writer states its intent explicitly; not-null-without-default
+-- forces the next one to do the same.
+ALTER TABLE public.sessions
+  ALTER COLUMN status SET NOT NULL;
+
+-- Added VALID in one step, not NOT VALID + VALIDATE. All 92 rows conform and the
+-- scan is trivial, so the two-step form would buy nothing and leave a second step
+-- to forget. (sessions_date_parseable is NOT VALID for the opposite reason - it
+-- was added over data that could not all be guaranteed to pass.)
+ALTER TABLE public.sessions
+  ADD CONSTRAINT sessions_status_known
+  CHECK (status IN ('actual', 'completed', 'logged', 'planned', 'cancelled'));
+
+COMMENT ON COLUMN public.sessions.status IS
+  'Lifecycle of the session. The canonical list lives in '
+  'web/src/constants/sessionStatus.js and MUST stay in step with the '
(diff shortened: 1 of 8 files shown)


skizzy1986:
**Release note draft**

**Fixed** — The Log and the "latest erg session" summary could show sessions out of order: a session logged or edited well after it happened would appear as your most recent one, burying genuinely newer sessions further down the list. Both now sort by the date you actually trained, so they reflect what really happened most recently. (Session status values are also now validated at the database level, closing off a way bad data could sneak in and cause this kind of thing.)

---
_Generated by [Claude Code](https://claude.ai/code/session_012QLfnwECDCZqT5bGvTYm41)_

---

PR #259, opened 2026-08-23
config: move @capacitor/cli to devDependencies

skizzy1986:
## What changed and why

`@capacitor/cli` was in `dependencies`. It is a build-time Node CLI — the package exposes only `bin: [capacitor, cap]`, nothing in `web/src/` imports it, and browser code cannot import it at all. It belongs alongside `vite`, `vitest` and `eslint`.

One line moved in `web/package.json`; the lockfile regenerated to carry the resulting `dev` flags.

### The concrete effect

`ci-web.yml:63` gates on `npm audit --omit=dev --audit-level=high`, and the CLI was pulling an advisory into that scope:

```
@capacitor/cli@8.5.0  →  xcode@3.0.1  →  uuid@7.0.3
```

[GHSA-w5hq-g745-h8pq](https://github.com/advisories/GHSA-w5hq-g745-h8pq) — moderate, `uuid < 11.1.1`. npm counts it three times, once per node of the chain.

| scope | before | after |
|---|---|---|
| `npm audit --omit=dev` — **what CI gates on** | 3 moderate | **0** |
| `npm audit` (all) | 3 moderate | 3 moderate — unchanged, now correctly dev-scoped |

**This is dependency hygiene, not a patch.** Worth being explicit, because the numbers could be misread as a fix:

- The advisory has **no upstream fix**. `@capacitor/cli@8.5.0` *is* latest and still pins `xcode ^3.0.1`, which still pins `uuid ^7.0.3` (uuid is on 14.x). The chain is stale upstream.
- `npm audit fix --force` would **downgrade** the CLI to 8.4.2 — a breaking change that does not resolve it either. Don't run it.
- The vulnerable code never shipped anyway: `uuid` is absent from `dist/` (no `byteToHex`, no NIL constant — every `uuid` string in the bundle is a Bluetooth GATT property from the BLE plugin), and `xcode` is iOS tooling in an Android-only project.

## The lockfile needed care

Regenerated with `npx npm@latest install --package-lock-only`, **not** this container's npm **10.9.7** — npm 10 strips the 10 `"libc"` blocks off the optional platform deps, which is the exact drift #241 was merged to stop. Doing it the lazy way would have reintroduced that inside a dependency-tidying PR.

The diff is 77 `"dev": true` + 2 `"devOptional": true` flags, plus the one relocated line in the root package entry. `lockfileVersion`, every `libc` block and every version are untouched — 80 insertions, 1 deletion.

## How to test manually

Not needed — no runtime surface changes. `vite build` never touches the CLI.

The one command that genuinely needs it is `npx cap sync android`, and `ci-android.yml` runs plain `npm ci` (not `--omit=dev`) in both jobs beforehand, so the CLI is still installed. The **Build Debug APK** job is the real check here.

## Checklist

- [x] Tests cover the change, or I've noted why they don't — dependency metadata only, no code changed; the full 720-test suite passes unchanged as a regression check
- [x] No unexpected console errors in the browser — no runtime surface touched

Also verified locally: `npm ci` accepts the npm12-written lockfile under npm 10.9.7 (matching CI's Node 22), and `npx cap --version` still resolves 8.5.0 from `node_modules/.bin`.

---
_Generated by [Claude Code](https://claude.ai/code/session_017qkwjR61snwGxwSgwLKf3W)_

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -10,7 +10,6 @@
       "dependencies": {
         "@capacitor/android": "^8.5.0",
         "@capacitor/app": "^8.1.1",
-        "@capacitor/cli": "^8.5.0",
         "@capacitor/core": "^8.4.2",
         "@capacitor/local-notifications": "^8.3.0",
         "@capacitor/network": "^8.0.0",
@@ -25,6 +24,7 @@
         "vite-plugin-pwa": "^1.3.0"
       },
       "devDependencies": {
+        "@capacitor/cli": "^8.5.0",
         "@eslint/js": "^10.0.1",
         "@playwright/test": "^1.62.1",
         "@sentry/vite-plugin": "^5.4.0",
@@ -1665,6 +1665,7 @@
       "version": "8.5.0",
       "resolved": "https://registry.npmjs.org/@capacitor/cli/-/cli-8.5.0.tgz",
       "integrity": "sha512-rLdzMUM5QV4WITcqoWv04p32i14BXgUH2diqEH6MWQlWaJfiyNrvOyt/+d5vHAfOxOm1klBu637VDEobctlwBA==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@ionic/cli-framework-output": "^2.2.8",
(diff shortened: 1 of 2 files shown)


skizzy1986:

This moves `@capacitor/cli` from `dependencies` to `devDependencies` (with the corresponding lockfile regeneration) to correctly scope it out of the `npm audit --omit=dev` CI gate. No runtime code changed, no user-facing behavior affected.

---

Issue #183, opened 2026-08-18
Design tokens: finish the THEME migration (slices 2c–2e)


skizzy1986:
Tracking issue for the remainder of the design-token consolidation. Slices 1, 2a and 2b have all shipped; there has never been a tracking issue for what follows (creation was blocked when 2b was specced).

## Shipped so far

| Slice | PR | Landed |
|---|---|---|
| 1 — `theme.js` + CSS vars + 8 `const C` blocks | #149 | `cfbc3b9` |
| 2a — 8 new tokens for high-frequency colours | #153 | `b8fda47` |
| 2b — 11 shared components (~60 literals) | #154 | `a2675a8` (2026-08-18) |

`web/src/constants/theme.js` now exports **23 tokens**; `web/src/utils/themeCss.js` projects them to `--color-*` custom properties injected at `:root` in `main.jsx`.

## What remains — measured on main, 2026-08-18

**1,204 hardcoded hex literals across 37 files.** The five big desktop views hold ~57% of them and **none of them import `THEME`**:

| File | Literals | Lines |
|---|---|---|
| `views/ProgramView.jsx` | 217 | 1,936 |
| `views/OverviewView.jsx` | 124 | 1,361 |
| `views/ErgView.jsx` | 121 | 1,235 |
| `views/program/ProgramYear.jsx` | 109 | 936 |
| `views/RecoveryView.jsx` | 86 | 781 |

Next tier: `MobileAnalytics` 38, `StrengthLogger` 38, `JournalView` 37, `MobilityView` 31, `StrengthView` 29, `MobileRecovery` 27, `trainingConfig.js` 23, **`App.jsx` 23 (no THEME import)**, `ProgramMicrocycle` 22, `CalendarView` 21, `program.js` 16, `formatting.js` 15, `LogView` 12, `ui.js` 9.

## Proposed slices

- [ ] **2c — desktop views.** `OverviewView`, `ErgView`, `RecoveryView`, `JournalView`, `MobilityView`, `StrengthView`, `CalendarView`, `LogView`, `PlanView`, `ErgLiveView`. Largest single win. Split further if the diff gets unreviewable.
- [ ] **2d — mobile views + `App.jsx`.** `MobileAnalytics`, `MobileRecovery`, `MobileStrength`, `MobileSessionLog`, `MobileApp`, plus `App.jsx`'s 23.
[...]

PR #263, opened 2026-08-24
feat: surface cancelled sessions on the calendar day strip

### LINKED ISSUE #242: feat: show on the calendar that a day had a cancelled session, not just a completed one

skizzy1986:
Closes #242

## What changed and why

The calendar day cell now appends a `⊘ N CANCELLED` marker, so a day where a prescription was dropped is distinguishable from one where it never existed — both when something else was completed that day (7 dates) and when nothing was (16 dates).

## The shape of the change

This is **additive, not a correctness fix**. Cancelled rows never reached `dayStatus`: every caller passes `loggedSessions`, which `useSessionLog` builds through `isCompletedStatus`. Those 7 days rendered `done` because something genuinely *was* completed, and the 16 rendered `missed` because nothing was logged. Both correct. What was missing was the other half of the story.

| state | with a cancellation |
|---|---|
| `✓ DONE ×2` | `✓ DONE ×2` `⊘ 1 CANCELLED` |
| `— not logged` | `— not logged` `⊘ 2 CANCELLED` |
| `● TODAY` | `● TODAY` `⊘ 2 CANCELLED` |
| `UPCOMING` | `UPCOMING` `⊘ 1 CANCELLED` |

A cancelled-only past day also lifts from `0.5` to `0.75` opacity, so a recorded decision no longer dims like an absence. That is the only behavioural change to an existing line.

The `upcoming` + cancelled case is real, not theoretical — 7/2/26 is `Lower B — DROPPED (retest taper)`, cancelled ahead of time during a taper.

## How the #194 contract is kept

`utils/schedule.js` is **not touched**. Its contract says callers must pass an already-filtered list, so the view calls the exported, status-blind `logEntriesForDate` a second time with a pre-filtered cancelled list. The single status check (`isCancelledStatus`) lives one layer up in the hook; `logEntriesForDate` stays a pure date-matcher unaware that one caller feeds it a cancelled-only list. #194's bug was a second status gate *at the point of consumption* — this is the inverse. Reusing the one unpadded `M/D/YY` key builder also means the completed and cancelled counts cannot drift apart on a date-formatting bug.

`statusColor` and `statusLabel` are byte-identical; the marker is a sibling span. That makes "zero-cancelled renders unchanged" a property of the code rather than something tests have to police — and it is why the 16 existing `CalendarView.test.jsx` tests pass with zero edits.

The `×N` suffix stays completed-only: it derives from `dayStatus(...).logged`, which never sees the cancelled list.

## How to test manually

Not required — covered by CI. The cancelled rows are dated 23 Jun – 20 Jul 2026, so none fall inside the calendar's today−3 → today+14 window; the feature **cannot be verified by looking at the calendar today**. That is why the tests are clock-pinned.

`CalendarView.dayStrip.test.jsx` pins the clock to 26 Jun 2026 — the one 17-day window holding all five cases as real dates: both-states with and without a `×N` suffix, cancelled-only past, cancelled-only future, and a clean control. A second block pins 24 Jun so today itself is a cancelled-only day.

## Notes for review

- **Zero hex delta.** `CalendarView.jsx` gains a `THEME` import for the marker only (`THEME.textSubtle`, matching the `CANCELLED` badge `LogEntry` already renders). Its four pre-existing raw-hex literals are deliberately left alone — that migration is #183's.
- The second commit extends the existing R9 prop-wiring guard. The `cancelledSessions = []` default means a dropped prop would silently render zero markers while every test still passed; verified by mutation that removing the prop from `App.jsx` now fails.

Coverage: `CalendarView.jsx` 100/100/90, `sessionStatus.js` 100/100/100, `useSessionLog.js` 100/100/89.47. 740 tests pass.

## Checklist

- [ ] Visual change: yes — a new `⊘ N CANCELLED` marker and one opacity value. Not currently observable in the running app (all affected dates are outside the render window), so no screenshot is attachable; the rendered DOM is asserted in `CalendarView.dayStrip.test.jsx` instead.
- [x] Stays inside the property boundary: colour hexes = #183 · box metrics = S6 ·
      type = S-1 (see DESIGN.md)

---
_Generated by [Claude Code](https://claude.ai/code/session_01KTN1v6sUvJJTDRRRcfHkBn)_

### DIFF
--- a/web/src/App.jsx
+++ b/web/src/App.jsx
@@ -173,6 +173,7 @@ export default function App() {
 
   const {
     loggedSessions,
+    cancelledSessions,
     logDisplaySessions,
     plannedSessions,
     loggedKeys,
@@ -372,7 +373,11 @@ export default function App() {
 
           {/* ── CALENDAR VIEW ── */}
           {view === 'calendar' && (
-            <CalendarView loggedSessions={loggedSessions} isWide={isWide} />
+            <CalendarView
+              loggedSessions={loggedSessions}
+              cancelledSessions={cancelledSessions}
+              isWide={isWide}
+            />
           )}
 
           {/* ── PROGRAM VIEW ── */}
--- a/web/src/__tests__/App.test.jsx
+++ b/web/src/__tests__/App.test.jsx
@@ -156,6 +156,9 @@ vi.mock('../views/CalendarView.jsx', () => ({
     <div>
       <div>CalendarView-stub</div>
       <div data-testid="calendar-counted">{props.loggedSessions?.length}</div>
+      <div data-testid="calendar-cancelled">
+        {props.cancelledSessions?.length}
+      </div>
     </div>
   ),
 }));
@@ -301,6 +304,7 @@ describe('App', () => {
 
     fireEvent.click(screen.getByRole('button', { name: 'CALENDAR' }));
     expect(screen.getByTestId('calendar-counted')).toHaveTextContent('2');
+    expect(screen.getByTestId('calendar-cancelled')).toHaveTextContent('1');
 
     fireEvent.click(screen.getByRole('button', { name: 'LOG' }));
     expect(screen.getByTestId('logview-shown')).toHaveTextContent('3');
--- a/web/src/constants/sessionStatus.js
+++ b/web/src/constants/sessionStatus.js
@@ -9,3 +9,8 @@ export const COMPLETED_STATUSES = ['actual', 'completed', 'logged'];
 // happened. Deliberately NOT in COMPLETED_STATUSES: it must never clear a
 // benchmark, only reschedule one.
 export const PLANNED_STATUS = 'planned';
+
+// A session that was decided against — a deviation from the plan, not training
+// that happened. Scalar rather than an array: unlike COMPLETED_STATUSES it has
+// exactly one spelling in the CHECK allow-list.
+export const CANCELLED_STATUS = 'cancelled';
--- a/web/src/hooks/__tests__/useSessionLog.test.js
+++ b/web/src/hooks/__tests__/useSessionLog.test.js
@@ -361,3 +361,77 @@ describe('useSessionLog — cancelled is shown but not counted (#194)', () => {
     expect(shownOnly.every((e) => e.status === 'cancelled')).toBe(true);
   });
 });
+
+// ── #242: cancelled carried as its own list for the day strip ─────
+// The calendar needs to mark a day where a session was cancelled without
+// re-inspecting `status` in the view. The list is a strict subset of the SHOWN
+// set and disjoint from the COUNTED set — that relationship is the contract.
+describe('useSessionLog — cancelled sessions are carried separately (#242)', () => {
+  it('AC1 puts cancelled session 61 in its own list, out of the counted one', async () => {
+    mockQuery([
+      session61,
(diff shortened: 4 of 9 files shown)


skizzy1986:
**Release note draft (New)**

The training calendar now flags cancelled sessions right on the day strip — a day where a planned workout was dropped shows a `⊘ N CANCELLED` tag alongside (or instead of) the usual done/missed/upcoming status. Previously a cancelled session left no visible trace, so a dropped workout looked identical to a day that was never scheduled; now you can tell the two apart at a glance, even weeks later.

---

PR #262, opened 2026-08-23
docs: bring the design workstream into the repo and open a design lane

skizzy1986:
## What changed and why

The design workstream produced a 688-line UI & IA brief, a light redesign spec, five finished screens and a CI guard — and referenced none of it from any top-level document, while the document that governs it had never been committed at all. This lands the record in git, makes it discoverable, fixes the places where the docs contradicted the code, and opens design as a tracked lane.

No `Closes #` — this PR is what produced the issue set (#248–#261) rather than closing one of them. It does carry the delivery of #257 (commit `983dcd9`), which was closed on receipt of the design drop.

---

### The governing document was not in the repo

`HANDOFF.md` is cited eight times as normative — it sets the light palette, the `var(--color-*)` token seam, the role rename and the Archivo weight-500 floor. `git log --all -- '*HANDOFF*'` was empty; only fragments survived, quoted second-hand in three files. It is now committed verbatim at the path it names for itself, alongside the five artboards, `ISSUES-load-states.md` and `PROJECT-CONTEXT.md` from the design project.

The artboards are self-contained — no network, no build step, no external reference. Double-click and they render. 3.9MB on disk, 2.66MB packed.

### Three things in the spec that do not survive contact with the code

Recorded in `NOTES.md` rather than edited into `HANDOFF.md`, which is a decision record:

1. **Its acceptance list asks someone to confirm that existing tests assert structure and not colour.** Confirmed false. Seven files lock colour — `theme.test.js` locks all 23 hex values and requires `/^#[0-9a-f]{6}$/`, a regex that forbids `var(--color-*)`; five more assert derived colours; and `e2e/smoke.spec.js:108` asserts `rgb(0, 212, 255)` across all 13 tabs. **Consequence: "all 11 components in one PR" becomes four bisectable steps** (#248 → #250 → #251, with #253/#252 alongside).
2. **The seam inverts a dependency it does not mention.** `cssVars(THEME)` *generates* the CSS variables from `THEME`; once `THEME.accent` is `'var(--color-accent)'` it emits a self-referential variable that resolves to nothing. `theme.js` has to split into a values module and a pointer table.
3. **Nine tokens have no light value**, and the ground colour is given twice with different answers.

### Two things the design drop overturned

- **Dark is retained as a second theme, not deleted.** `HANDOFF.md` §1 says "dropped"; the project's `conventions.md`, revised a day later, overturns that in review — the erg room is dark at 5am. #251 corrected; it previously instructed deleting the dark values.
[...]

Issue #258, opened 2026-08-23
chore(design): re-sync and open the SplitIQ design project — owner action


skizzy1986:
The uploaded design system is stale and the project **has never been opened**. `STATE_OF_PLAY.md` §1 proves the second part positively rather than by absence: `_ds_needs_recompile` is still present and still ours (`{"by":"design-sync-cli"}`), and `_ds_manifest.json` is absent — the app writes the manifest and clears the sentinel the first time anyone opens the project.

## What the project currently holds

Built at `6dae5c1`, uploaded 2026-08-21, untouched since:

- The **old dark** `conventions.md` — the repo's copy has moved on three times (#224, #240, and the reconciliation)
- A **live, stale `PACE_ZONES`** frozen from the old seed CP of 190. `NOTES.md` is blunt about which half is worse: *"A dead export fails loudly; a stale-but-live constant quietly hands out wrong pace bands."* Until re-synced, that is what the design agent loads.

## Do not re-sync yet

Two prerequisites, in order:

1. **The four-way reconciliation must land first.** #240 rewrote `conventions.md` to light/role tokens but left 15 colour-named references in `docs/*.md`, `previews/*.tsx` and `dtsPropsFor`, so the agent receives both systems in one prompt. A status banner now says which is live, but the references themselves move with the role rename (#248). Re-syncing before that just freezes the contradiction into another bundle.
2. **Re-sync again immediately after the rename** — it changes every prop contract the agent codes against.

---

## Upload two new documents to the project root

`web/.design-sync/CLAUDE.md` and `web/.design-sync/CODE-TO-DESIGN.md` exist in git but **a design session cannot read the repo** — the inverse of the constraint `HANDOFF.md` opens with. They only become reachable once uploaded.

`conventions.md` (the `readmeHeader`, and the only input guaranteed to be inlined into the agent's prompt) now points at both **by project-root path**. Until they are uploaded, those are dangling pointers.

**Do not put them in `docsDir`.** That directory becomes `components/<group>/<Name>/<Name>.prompt.md`, keyed by `category:` frontmatter — a non-component doc dropped there would surface as a phantom component in the Design System pane. They need to land at the project root alongside the README.

If the sync tooling has no mechanism for arbitrary root-level files, the fallback is to concatenate `CLAUDE.md` ahead of `conventions.md` in whatever feeds `readmeHeader`. Less clean — it re-merges two documents that were deliberately split — but a guaranteed read beats an unreachable file.

[...]

PR #240, opened 2026-08-23
docs: reconcile the design-sync surface against HANDOFF.md

skizzy1986:
Related: #183, #201, #225. No `Closes` — #183 is reframed here, not closed.

> **Rewritten after rebasing onto main.** This PR originally also repaired the
> design-sync barrel. #225 and the `fix/design-sync-barrel-and-contrast` work landed
> on main first, so all of that is dropped — main's version was taken for every
> barrel file. What remains is only the reconciliation, which nothing else covers.

## What changed and why

The synced design surface still described a **dark-only** system, with a typeface and
a theming mechanism that `HANDOFF.md` had already superseded. Two decisions were
recorded in `DESIGN_BRIEF.md` §8.1 as *Resolved* that `HANDOFF.md` contradicts —
worse than leaving them open, because a reader trusts them.

- **Typeface** — recorded as IBM Plex Sans + Mono. It is **Archivo, minimum weight
  500**; below that it fails on light grounds at the sizes this app uses, so any
  component still setting `fontWeight: 400` has to be raised. Corrected in the brief
  and in the two `NOTES.md` bullets that named Plex.
- **The seam** — recorded as a `style`-prop passthrough on ten layout primitives.
  `HANDOFF.md` §1 reaches theming a different way: `THEME` values become
  `var(--color-*)` strings and a `data-theme` attribute redefines them beneath the
  root, so component source keeps its shape (`color: THEME.accent` still works) and
  the alias maps survive untouched. §1.8's *diagnosis* stands; its §5.1 prescription
  is not what is being built. Recorded as **superseded, not built** — and noting that
  a `style` seam may still be wanted for *layout composition*, which the token seam
  does not solve, so that question is reopened rather than closed either way.

**Dark is dropped; light is the only theme.** `conventions.md`'s theme sections are
rewritten against the light palette: role-named tokens (`cyan`→`accent`,
`gold`→`caution`, `pink`→`accentAlt2`, …), the weight-500 floor, z-index layering
(sheet 200 / backdrop 150 / nav 100), definite track heights, and the chart rules
from `HANDOFF.md` §3 — including *never use the min/max envelope of the window being
plotted* as a reference band, and *any caption stating a count or rank is computed,
never typed*.

**The contrast finding inverted.** The `muted`-fails-AA defect dissolves with the
dark palette — light `muted` is `#43485a` on white at 9.08:1. The live constraint is
its mirror image: every role token clears AA on white `surface` and **fails on the
`bg` ground**.

[...]

PR #266, opened 2026-08-24
fix(design): verify conventions.md's contrast claims, and correct two stale docs

skizzy1986:
Refs #258, #262 — deliberately not `Closes`. #258's acceptance criteria all live on claude.ai/design and need an interactive `/design-login`; this makes a future re-sync safer, it does not perform one.

## What changed and why

Three corrections found while working #258, kept separate from #262 which landed the design workstream itself.

**1. `conventions.md`'s contrast claims are now recomputed, not trusted.**

`conventions.md` is the `readmeHeader` — the one input guaranteed to be inlined into the design agent's prompt — so a wrong ratio in it reaches every generated design. Its own revision history is the argument for checking it rather than trusting it:

- #224 measured the ratios
- #240 replaced them with a different palette's
- #262's commit message records that *"four separate AA failures were shipped during this revision, every one of them a colour chosen against a background that was darkened afterwards"*

Ratios in prose do not recompute themselves. `npm run check:design-sync` now verifies:

- every `N:1+ against #hex` floor survives arithmetic
- every exact `#fg on #bg -> N.NN` recomputes
- every `--color-*` named exists in `THEME`, or it resolves to nothing in a design

The success line prints how many claims it verified, so a pattern that stops matching shows up as a count drop rather than a silent pass. Today: `1 published contrast claim(s) in conventions.md recompute.`

**2. `STATE_OF_PLAY.md` listed both §2 defects as live and "blocks any re-sync".** Neither is:

| Defect | Status |
|---|---|
| §2.1 barrel / build broken | Fixed in `85eab51` (#224), guard in #225. Re-verified at `ad2a8bd` — bundle builds, `check:design-sync` passes |
| §2.2 contrast failure | Fixed in #224 → **reverted by `e61092c` (#240)** → closed again by #262, which drops the `THEME.muted` recommendation for `#43485a` (5.26 on the light ground, 9.08 on a white card, both pass AA) |

**3. `CLAUDE.md` gave the AT band as `164–205 W`.** `derivePaceZones(205)` yields **164–185**; 205 is the CP anchor itself, not AT's ceiling. UT2 and UT1 were already correct.

**4. `lint` and `format:check` covered `src/` only**, so the CI-enforced design-sync guard was itself ungated. Both now include `scripts/`, with Node globals for `scripts/**/*.mjs`.

## How to test manually

```
npm run check:design-sync   # 1 claim verified
```

Negative-tested all three checks — each fails as expected:

| Mutation | Result |
|---|---|
| Falsify the AA floor (darken the ground) | `claims #43485a clears 4.5:1 against #5a6070; it is 1.44` |
| Publish a wrong exact ratio | `says #7e7e9a on #2a2a48 is 9.99; it is 3.50` |
| Name `--color-accent` (not in `THEME`) | `theme.js does not define — it will resolve to nothing` |

## Checklist

- [x] Tests cover the change, or I've noted why they don't — the guard is verified by mutation rather than by unit test, matching how `check-design-sync-entry.mjs` was already built; no app source under `web/src/` is touched, so the 720-test suite is a regression check here

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -234,7 +234,7 @@ legacy `sessions.date` text pattern); RLS single-owner policy like the modern ta
 - **sRPE** — How hard a session felt on a 1–10 scale (subjective).
 - **CP** (Critical Power) — The highest power you can sustain indefinitely.
   **~205W provisional (rowing)**; revalidate via rested 1-min + 4-min tests.
-  Rowing zones off this anchor: **UT2 113–144 / UT1 144–164 / AT 164–205 W**.
+  Rowing zones off this anchor: **UT2 113–144 / UT1 144–164 / AT 164–185 W**.
 - **Current model — pure base + strength** (reverted from polarised on 2026-06-29):
   rowing is aerobic volume only (UT1/UT2 — no programmed threshold/VO₂); the bike
   is a complementary Z1/Z2 aerobic carrier, never a programmed intensity source;
--- a/web/.design-sync/STATE_OF_PLAY.md
+++ b/web/.design-sync/STATE_OF_PLAY.md
@@ -7,8 +7,11 @@ neither — it is a status check, and it goes stale quickly._
 Rendered version: https://claude.ai/code/artifact/e16750c2-4d67-4bc2-99a3-7fae8ca450dc
 
 > **Verdict.** The Claude Design project has never been opened. Separately, `main`
-> moved underneath the synced design system and the design-sync build is now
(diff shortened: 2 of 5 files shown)


skizzy1986:
Internal change — no release note. This PR only touches docs (`CLAUDE.md`, `STATE_OF_PLAY.md`) and CI/lint tooling (`eslint.config.js`, `package.json` script scope, `check-design-sync-entry.mjs`); no `web/src` app code changed, so there's nothing user-visible to log.

---

PR #267, opened 2026-08-24
test: screenshot baselines for the app shell (S6.0)

### LINKED ISSUE #249: test: screenshot baselines for the app shell (S6.0)
The brief queues "S6.0 = screenshot baselines first" before any pixel-moving slice, and there is no visual-regression net today — one Playwright smoke spec with no screenshot assertions.

Without this, the palette rename cannot *prove* it changed no pixels, and the spacing pass is a ~200-site change with nothing catching drift.

## Shape

Reuse what exists. **No new dependency, no new service** — `web/e2e/smoke.spec.js` already solves the hard part (getting past the Supabase auth gate by seeding `localStorage` and stubbing backend traffic).

- Extract that `beforeEach` into `web/e2e/fixtures.js`; the smoke spec keeps using it, no behaviour change.
- New `web/e2e/visual.spec.js` using Playwright's built-in `toHaveScreenshot()`.
- `npm run test:visual` script.
- Fixed viewports: **1280×800 desktop and 390×844 mobile**. The mobile shot is not optional — the 767px fork is a separate app, so without it half the product is uncovered.
- `reducedMotion: 'reduce'`, `deviceScaleFactor: 1`, `animations: 'disabled'`.

## Two constraints that decide whether this survives

- **Mask the charts.** `NOTES.md` records that a pinned clock stalls Recharts' `react-smooth` line-draw mid-animation, and it cannot be disabled from outside because `PaceTrendChart` hardcodes `<Line>` without `isAnimationActive`. Mask chart containers in v1 and say so in the spec header — S6 cannot lean on this for chart layout.
- **Baselines must be generated in the CI Playwright container**, never locally. Glyph rasterisation differs; a locally-generated baseline fails forever. This is the most common way these harnesses die.

## Scope — 6 to 10 shots

Baseline the **shell and chrome**: nav, tab bar, headers, section labels, `ErrorFallback`, and 2–3 representative views. The stubs return `[]`, so data-dense interiors render as empty states — stable but low-value. Content fixtures are a later issue; don't pretend coverage exists.

## CI

Second job in `.github/workflows/e2e-web.yml` — same container, same lockfile-pinned version. Land it **non-required**, uploading the diff artifact. Promote to required only after it has been quiet across several PRs; this repo has already been deadlocked for a month by an over-eager gate.

## Acceptance

- [ ] Two consecutive runs on an unchanged tree produce byte-identical output
- [ ] Nudging one `padding` value fails with a readable diff artifact
- [ ] Baselines committed from a CI run, not a dev machine

Blocks the palette rename and the Archivo swap.

skizzy1986:
Closes #249

## What changed and why

Adds a visual-regression net — nine `toHaveScreenshot()` baselines for the app shell — so the palette rename can prove it moved no pixels and the S6 spacing pass has something that catches drift. No new dependency, no new service: the auth-gate and backend-stubbing setup already solved in `smoke.spec.js` moves verbatim into `web/e2e/fixtures.js` and both specs share it.

**Nine shots**, one `visual` project:

| Viewport | Shots |
|---|---|
| 1280×800 | `desktop-overview`, `desktop-calendar`, `desktop-program`, `desktop-journal`, `error-fallback`, `login` |
| 390×844 | `mobile-analytics`, `mobile-log`, `mobile-recovery` |

The mobile shots are not optional — `main.jsx:289` forks on a 767px media query into an entirely separate component tree.

**Three things that decide whether this survives:**

- **Charts are masked.** `.design-sync/NOTES.md` records that a pinned clock stalls Recharts' `react-smooth` line-draw part-way, and that it can't be disabled from outside because `PaceTrendChart` hardcodes `<Line>` without `isAnimationActive`. Every `.recharts-responsive-container` is masked, so **S6 cannot lean on these baselines for chart layout** — though the mask tracks the container's bounding box, so container geometry is still covered.
- **Baselines came from the CI container**, `mcr.microsoft.com/playwright:v1.62.1-noble` — same image, tag and architecture the `Playwright Visual` job runs. The default snapshot path template keeps the `-linux` suffix, so an accidental dev-machine baseline lands under a different filename instead of silently replacing the real one.
- **The clock is pinned**, which #249 didn't ask for but the app requires: `OverviewView`, `CalendarView`, `MobileAnalytics` and `MobileRecovery` all render `new Date()`, so unpinned baselines would rot at midnight. Locale and timezone are pinned for the same reason. The pinned clock and the chart mask are complementary rather than competing — the clock is what makes baselines stable across days, the mask is what makes its one known side effect harmless.

Tab switching settles on *"the rendered tree changed"* rather than on active-tab styling, because that styling is a colour on desktop and a `font-weight` on mobile — both owned by other design workstreams. The harness shouldn't break when #183 or S-1 lands.

**CI:** a second **non-required** job in the same lockfile-pinned container, uploading `web/test-results/` (actual/expected/diff triplets) under a name distinct from the smoke job's, since `upload-artifact` rejects duplicates and both jobs write `playwright-report/`. Non-required is deliberate and matches the `npm audit` note in `ci-web.yml` — an over-eager gate already deadlocked this repo for a month. A `workflow_dispatch` **Update Visual Baselines** job makes the pure-CI regeneration route available for the palette rename, the Archivo swap, and any Dependabot bump of `@playwright/test`.

## How to test manually

```bash
docker run --rm -v "$PWD":/repo -w /repo/web -e HOME=/root -e CI=1 \
  mcr.microsoft.com/playwright:v1.62.1-noble \
  sh -c "npm ci && npm run test:visual && npm run test:e2e"
```

## Acceptance criteria (#249)

**1. Two consecutive runs on an unchanged tree produce byte-identical output** — verified in the container; all nine baselines unchanged by md5 across both runs.

```
9 passed (15.8s)
===== SECOND RUN =====
9 passed (12.9s)
BASELINES UNCHANGED across both verification runs
```

**2. Nudging one `padding` value fails with a readable diff artifact** — `App.jsx:336` nav padding `8px 6px` → `9px 6px`. Exactly the four desktop shots that render the nav failed; the five that don't render it passed. Reverted; `git diff src/` is empty.

```
✘ desktop shell › overview    21553 pixels (ratio 0.03) are different
✘ desktop shell › calendar    24635 pixels (ratio 0.03) are different
✘ desktop shell › program     22760 pixels (ratio 0.03) are different
✘ desktop shell › journal     52949 pixels (ratio 0.06) are different
✓ error fallback   ✓ mobile analytics   ✓ mobile log   ✓ mobile recovery   ✓ login
4 failed, 5 passed
```

The diff artifact renders the 2px shift below the nav in red — legible at a glance.

**3. Baselines committed from a CI run, not a dev machine** — generated in the CI image as above, never from a host browser.

## Checklist

- [x] CI is green (lint, test, build) — lint, format:check, check:design-sync, 720 unit tests, coverage, build and bundle size (390.1 KB / 400 KB) all pass locally
- [x] Tests cover the change, or I've noted why they don't — this PR *is* the test; smoke stays at 14 passed, confirming the fixtures extraction changed no behaviour
- [x] Visual change: **none** — this PR touches no file under `web/src/`; it only records what the app looks like today
- [x] Stays inside the property boundary: touches no colour hex, no box metric, no type property

## Scope — what this does not cover

The stubs return `[]`, so every interior renders as an empty state; content fixtures are a later issue. Chart internals are masked out by construction. Uncovered: desktop `plan`/`live`/`erg`/`strength`/`logger`/`mobility`/`recovery`/`log`/`coach`, mobile `erg`/`coach`, `StrengthLogger` entirely, and every hover, focus and expanded state. This is stated in the spec header too, so nobody reads it as more coverage than it is.

---
_Generated by [Claude Code](https://claude.ai/code/session_014cc694sEGRtGSvHej6vnW9)_


skizzy1986:
Internal change — no release note. This PR adds Playwright visual-regression baselines and CI tooling only; it touches no file under `web/src/` and has no user-visible effect.

---

PR #268, opened 2026-08-24
fix(design): correct the AT band in the design briefing, and check it mechanically

skizzy1986:
Refs #258, #262, #266.

## What changed and why

#266 fixed the AT zone band in the repo-root `CLAUDE.md`. #262 landed `web/.design-sync/CLAUDE.md` in parallel carrying **the same error**, so the wrong number survived in the worse of the two places — the file a design session reads to learn what the numbers mean, and which is uploaded to the design project.

Both said AT is `164–205 W`. `derivePaceZones(205)` yields **164–185**:

```
UT2 113-144   ✅ was already right
UT1 144-164   ✅ was already right
AT  164-185   ❌ documented as 164-205
```

In both files the CP anchor (`~205 W`) sits one line above, which is how the ceiling got restated as the band's top.

Fixing the second copy by hand would leave the same gap that let this through twice, so it's checked instead. `npm run check:design-sync` now recomputes any UT2/UT1/AT watt bands published in the design-sync docs against `derivePaceZones`, at the CP the document itself quotes.

Same argument as the contrast check in #266: a band derived from a provisional anchor is a snapshot, and snapshots don't recompute themselves. `NOTES.md` already records the version of this that shipped — a bundle frozen at an old CP that *"quietly hands out wrong pace bands."*

## How to test manually

```
npm run check:design-sync
```

Reproduced the shipped bug against the new check:

| Mutation | Result |
|---|---|
| Reintroduce `164–205` | `gives AT as 164-205 W; derivePaceZones(205) yields 164-185` |
| Wrong UT2 floor (`100–144`) | `gives UT2 as 100-144 W; derivePaceZones(205) yields 113-144` |
| Clean tree | passes |

## Checklist

- [x] Tests cover the change, or I've noted why they don't — the guard is verified by mutation, matching how `check-design-sync-entry.mjs` was already built; no app source under `web/src/` is touched, so the 740-test suite is a regression check here

---

### DIFF
--- a/web/.design-sync/CLAUDE.md
+++ b/web/.design-sync/CLAUDE.md
@@ -54,7 +54,7 @@ Terms that appear on screen. Get these wrong and the design is wrong.
   types after a session.
 - **CP** — critical power, the wattage sustainable more or less indefinitely. ~205W,
   provisional. Every rowing zone derives from it.
-- **UT2 / UT1 / AT** — aerobic training bands off CP: 113–144 / 144–164 / 164–205 W.
+- **UT2 / UT1 / AT** — aerobic training bands off CP: 113–144 / 144–164 / 164–185 W.
   Almost all rowing volume is UT1/UT2; there is no programmed threshold work right now.
 - **Microcycle** — one week's pattern. **Home weeks** load, **FIFO weeks** deload.
 - **Readiness** — a derived morning score from RHR, HRV and sleep against personal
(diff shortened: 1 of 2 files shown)


skizzy1986:
Internal change — no release note. This corrects a documentation error in `web/.design-sync/CLAUDE.md` (the AT training-zone band) and adds a mechanical check script; no application code under `web/src/` changed, so there's no user-visible behavior to note.

---

PR #269, opened 2026-08-24
fix(design): give every .design-sync file one owner, and stop the two-way flow

skizzy1986:
Refs #258, #262.

## What changed and why

`conventions.md` moved in **both directions** with nothing arbitrating it. Three documents disagreed about who owns it:

| Source | Says |
|---|---|
| `config.json:58` | `readmeHeader: .design-sync/conventions.md` — pushes the **repo's** copy **up** |
| `PROJECT-CONTEXT.md` | project is "✅ source of truth", repo is a "mirror" — flows **down** |
| `HANDOFF.md:176` | §3 titled *"Replacement for `web/.design-sync/conventions.md`"* — design **authors** it |

Whichever side acted last silently won. That's not hypothetical — it's how `e61092c` (#240) replaced `85eab51` (#224)'s measured dark contrast ratios with the light palette's, in the file that was then guaranteed to reach the design agent's prompt. A re-sync today would have pushed the repo's copy back over whatever the design side has changed since.

`NOTES.md`'s "Sync inputs this repo owns" table didn't even list `conventions.md` — the one file the repo was actually pushing.

### The fix

`ownership.json` declares the direction per path, and `check:design-sync` enforces it:

| Owner | Meaning |
|---|---|
| `repo` | written here, pushed to the design project |
| `design` | authored in the project; **read-only mirror** here so a code session can read it |
| `local` | working docs — neither pushed nor mirrored |

Three checks:
- every file in `.design-sync/` is classified **exactly once**
- no manifest entry matches nothing (stale entries fail)
- **nothing `config.json` uploads may be design-owned** ← this is the fix; the rest keeps the manifest honest

### `readmeHeader` now points at `CLAUDE.md`

The repo needs *something* as the design system's README, and `conventions.md` was the wrong choice because the repo doesn't own it. `CLAUDE.md` — the domain briefing — is the right one, because the domain genuinely lives in the code (its own header says so).

`CODE-TO-DESIGN.md` had already asked for this: *"If it does not, point `readmeHeader` at it."* It also means the synced design system stops shipping a second, stale copy of the style guide — the "Known tension" `PROJECT-CONTEXT.md` records. And it resolves `CLAUDE.md` having had **no wiring at all**, which was the original item.

## How to test manually

```
npm run check:design-sync
# → ...1 published contrast claim(s) recompute, and all 42 files in
#   .design-sync/ have a declared owner with nothing design-owned being pushed back.
```

Verified by mutation:

| Mutation | Result |
|---|---|
| Repoint `readmeHeader` at `conventions.md` (the original bug) | `...which ownership.json says the design project owns. Pushing it would overwrite the source of truth — this is the loop that lost #224's contrast ratios.` |
| Unclassified file appears | `does not classify .design-sync/SCRATCH.md — declare it repo, design or local` |
| Manifest lists something gone | `lists "GONE.md", which matches no file in .design-sync/` |

## Not done on purpose

**Design-owned files are untouched.** `PROJECT-CONTEXT.md` still describes the old arrangement and `conventions.md` still carries claims that follow from it — correcting those is a change to make **on the design side**, not in a mirror this manifest now declares read-only. Editing them here would be the same clobber this PR exists to prevent.

## Checklist

- [x] Tests cover the change, or I've noted why they don't — the guard is verified by mutation, matching how `check-design-sync-entry.mjs` was already built; no app source under `web/src/` is touched, so the 740-test suite is a regression check here

---

### DIFF
--- a/web/.design-sync/CODE-TO-DESIGN.md
+++ b/web/.design-sync/CODE-TO-DESIGN.md
@@ -130,10 +130,14 @@ freeze the contradiction into another bundle. **Re-sync after the role rename la
 (#248), which is when those references move in lockstep with the code. Re-run
 `node .design-sync/gen-css.mjs` first — step one of every sync. Tracked as #258.
 
-While re-syncing, verify how `CLAUDE.md` reaches the design agent. `readmeHeader` inlining
-is the only channel *confirmed* to work; whether the app auto-loads a file by that name is
-unverified. If it does not, point `readmeHeader` at it. The behavioural test: ask for a
-chart and see whether it produces a labelled axis and a computed caption unprompted.
+**Done — `readmeHeader` now points at `CLAUDE.md`.** This section asked for it, and it
+was the right call for a second reason: `readmeHeader` previously pushed `conventions.md`
+*up*, over the copy the design project owns. That was a two-way flow with nothing
+arbitrating it. `ownership.json` now declares the direction per file and
+`npm run check:design-sync` fails if `config.json` ever names a design-owned path again.
+
+The behavioural test still stands: ask for a chart and see whether it produces a labelled
+axis and a computed caption unprompted.
 
 ## Build order — when the designs become buildable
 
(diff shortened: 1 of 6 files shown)


skizzy1986:
Internal change — no release note. This PR only touches the `web/.design-sync/` sync tooling (an ownership manifest + guard for which files flow which direction between the repo and the design project) and its own docs. No `web/src/` app code changed, so there's nothing user-visible to log.

---

PR #271, opened 2026-08-24
feat(observability): make Sentry the testing and auditing rail

### LINKED ISSUE #270: feat: make Sentry the testing and auditing rail — it has never reported an event
## What + why

`@sentry/react` has been a shipped dependency, `Sentry.init` runs at boot, and
`Sentry.ErrorBoundary` wraps the tree — but **Sentry has never received a single
event**. The org `splitiq-29` had zero projects, zero DSNs, zero issues.

Five defects kept it dark, each independently sufficient:

1. **No project, so no DSN.** `initSentry()` returns `false` in every environment
   including production.
2. **CSP blocks ingest.** `connect-src` in `web/vercel.json` lists only Supabase.
   Every envelope is refused by the browser before it leaves the page, DSN or not.
3. **The per-tab `ErrorBoundary` swallows render errors.** `web/src/App.jsx` has
   `getDerivedStateFromError` but no `componentDidCatch`, and it wraps the whole
   view switch — so nearly every render error in the app is caught, rendered as an
   inline panel, and never reported.
4. **`tracesSampleRate: 0.1` is inert.** `browserTracingIntegration` is not one of
   the browser SDK's default integrations, so nothing is ever sampled.
5. **`VITE_SENTRY_RELEASE` is read but set nowhere**, and the vite plugin passes no
   `release`, so the two would disagree — which makes Sentry serve minified frames
   without erroring.

Beyond the frontend, two gaps matter as much:

- **Supabase failures are invisible.** Eleven hooks `throw error` into react-query
  and nothing listens — no `QueryCache` `onError` anywhere. A broken query is
  silent.
- **The four edge functions have no telemetry at all**, and their Deno tests
  (`vitals-import/test.ts`, `vitals-import-api/test.ts`, 63 assertions) are run by
  **no workflow**.

## Acceptance criteria

- [ ] Sentry projects exist and events arrive with original (non-minified) frames
- [ ] The ingest host is allowed by CSP
- [ ] Render errors in a tab are isolated **and** reported
- [ ] Supabase failures reach Sentry
- [ ] Release name matches between the SDK and the uploaded artifacts
- [ ] Preview and production are distinguishable in the issue stream
- [ ] Vitest results feed Sentry Test Analytics for flaky-test detection
- [ ] The orphaned Deno tests run in CI
- [ ] Edge function failures and `vitals-import` cron liveness report
- [ ] A repeatable way to turn the issue stream into backlog items

## Out of scope

- Session Replay — costs ~35–50 KB gzip against ~10 KB of headroom under the
  400 KB budget in `web/scripts/check-bundle-size.mjs`.
- Edge-function deploy-drift detection — that is #210.

skizzy1986:
Closes #270

## What changed and why

Sentry was already integrated in code but had **never sent a single event** — five independent defects kept it dark, and two large blind spots (Supabase failures, edge functions) had no telemetry at all. Five commits, each standing on its own.

### 1. Make it report

- **CSP** (`web/vercel.json`): `connect-src` had no Sentry ingest host, so every envelope was refused by the browser. Adds the EU-region host — the org is EU-region, `*.ingest.de.sentry.io`.
- **Tracing** (`web/src/utils/sentry.js`): `browserTracingIntegration` is *not* a default integration, so `tracesSampleRate: 0.1` sampled nothing. Adds it, with `tracePropagationTargets` scoped to Supabase and same-origin so trace headers don't leak to third parties. A test asserts this so the config can't silently revert to inert.
- **The swallowing boundary** (`web/src/App.jsx`): the per-tab `ErrorBoundary` had `getDerivedStateFromError` but no `componentDidCatch`, so nearly every render error in the app was caught and never reported. Adds reporting; the isolate-the-tab behaviour is unchanged.
- **Supabase failures** (`web/src/utils/queryErrorHandlers.js`, new): eleven hooks throw into react-query and nothing listened. Adds `QueryCache`/`MutationCache` handlers, in a separate covered module because `main.jsx` is coverage-excluded. `useSessionLog` bypasses react-query entirely, so it captures its own.
- **Release** (`web/vite.config.js`): `VITE_SENTRY_RELEASE` was read but set nowhere. Derives it from Vercel's commit SHA and passes the **same string** to the vite plugin — when those diverge Sentry serves minified frames without erroring. Also sets `url: https://de.sentry.io`, since the plugin defaults to `sentry.io` and would not find an EU-region org.

### 2. Feed CI into Sentry

- Vitest emits JUnit XML alongside its normal reporter; the Test job uploads it via `getsentry/prevent-action` for flaky-test detection. Runs under `if: always()` (a failing run is the one worth uploading) and uses OIDC, so no token is stored. `fail_ci_if_error: false` deliberately — this is a reporting sink, not a gate.
- `supabase/functions/{vitals-import,vitals-import-api}/test.ts` existed but were run by **no workflow** — 63 assertions covering the CSV parser, the Google Health mapper and the timing-safe cron guard, none of them executing on any PR. Adds `ci-functions.yml`.

### 3. Cover the backend

Adds `supabase/functions/_shared/sentry.ts`, reporting to a separate `erg-dashboard-functions` project. Two things in it are required for correctness, not tuning: `defaultIntegrations: false` (the Deno SDK doesn't instrument `Deno.serve`, and the runtime reuses an isolate across requests, so global context would leak into the *next* caller's event) and `await flush()` before returning (the isolate can be frozen the moment the Response resolves).

`coach-chat`'s fetch to `api.anthropic.com` had no try/catch at all — a network failure escaped as an opaque platform 500. `vitals-import` additionally emits cron check-ins, started *after* the `CRON_SECRET` guard so an unauthorised probe isn't counted as a run.

### 4. Close the loop

Adds `/audit`: reads the Sentry issue stream over MCP, ranks by blast radius, verifies each candidate against the actual code, and files GitHub Issues behind an approval gate.

### 5. Environment labelling

`vite build` sets `MODE=production` for every build, Vercel target included, so preview deploys would have filed under `production`. Derives the environment from `VERCEL_ENV` instead, falling back to `MODE`.

## How to test manually

CI does not cover the parts that need real credentials. After setting `VITE_SENTRY_DSN`, `SENTRY_ORG`, `SENTRY_PROJECT`, `SENTRY_URL` and `SENTRY_AUTH_TOKEN` in Vercel (Production **and** Preview), on the preview URL:

1. Confirm a `POST` to `*.ingest.de.sentry.io/.../envelope/` returns 200 and is not blocked by CSP.
2. Trigger a render error in a tab and a failing Supabase query — both should appear in `erg-dashboard`.
3. Confirm the `environment` tag reads `preview`, not `production`.
4. **Confirm the stack trace shows original source, not minified frames.** This is the only real proof the release string and uploaded artifacts agree, and it is the one failure Sentry will not announce.

Edge functions also need `SENTRY_DSN` as a Supabase Edge Function secret (different project, different DSN).

## What was verified here, and what wasn't

Verified locally: the old CSP produces `Refused to connect` and the new one reaches the network (real headers served to Chromium); `browserTracingIntegration` is absent from the SDK's defaults (read from `node_modules`, not docs); the boundary test fails without `componentDidCatch`; `VERCEL_ENV=preview` yields ``environment:`preview` `` in the built bundle; all 63 Deno assertions pass under deno 2.9.5 and an injected failure exits 1.

**Not verified: no event has actually reached Sentry.** The agent sandbox's egress policy blocks `ingest.de.sentry.io`, so the final hop is untested. That is what steps 1–4 above are for.

## Checklist

- [x] CI is green — all 23 checks pass on `640f524`; **Deno Tests** green on its first ever run
- [x] Tests cover the change — 753 passing, up from 751; new per-file coverage gate on `queryErrorHandlers.js`
- [x] `npm run build` passes locally — 390.4 KB gzipped, 9.6 KB headroom. CI's bundle-size comment independently reports Δ ±0.0 KB vs main
- [x] Visual change: none
- [x] Stays inside the property boundary — `App.jsx` gains `componentDidCatch` only, no style changes

## Notes for review

- **Session Replay is deliberately excluded.** It costs ~35–50 KB gzip and there are 9.6 KB of headroom. Measured, not estimated.
- The free plan allows **one** cron monitor; it is on `vitals-import`, the job whose silent failure makes vitals stale and every readiness and load number downstream drift.
- **Test Analytics is wired but not yet receiving.** The first run's log resolves the open question from the original description: OIDC authenticated and the JUnit XML uploaded, then Sentry rejected it —
  ```
  info  -- Found 1 test_results files to report
  info  -- Upload queued for processing complete
  error -- Upload queued for processing failed: {"message":"Repository not found"}
  ```
  The repo is not registered with Prevent. `fail_ci_if_error: false` did its job: the step reported success and the PR stayed green rather than failing on an unprovisioned account. **No code change needed** — connecting the repo to `splitiq-29` via Sentry's GitHub integration will start it reporting, and the same integration is what Seer Code Review needs.

https://claude.ai/code/session_018rb9ttpvUKijJRP1DtZfjM


skizzy1986:
Internal change — no release note. This PR wires up observability (Sentry error/perf reporting, CI test-analytics upload, edge-function error handling) and adds the `/audit` skill; it fixes bugs that were silently swallowing telemetry but makes no user-visible change to the app itself ("Visual change: none" per the PR description).

---

PR #275, opened 2026-08-24
fix(coach): correct the AT band in the handover, and guard every published zone band

skizzy1986:
Refs #258, #266, #268.

Two commits: fix the third copy of a wrong zone band, then close the class so there isn't a fourth.

---

## 1. The handover corrections

`coach/CODE_TO_COACH_HANDOVER.md` is dated (2026-07-01), but its header says *"Paste into Coach chat when you want the two interfaces re-synced."* So stale facts in it reach a **live coaching context** rather than sitting harmlessly in a log.

Three fixes, each **annotated inline** rather than silently rewritten, since the document records what was done at the time:

| | Was | Now |
|---|---|---|
| AT band | `164–205 W` | `164–185 W` — `derivePaceZones(205)` yields this; 205 is the CP anchor, restated as the band's ceiling |
| Ordering | `to_date(...,'MM/DD/YY')` | order by generated `date_iso` — `CLAUDE.md:216` forbids `to_date` outright (it's STABLE, so it can't be indexed or used in a generated column) |
| Date format | `MM/DD/YY` | `M/D/YY` unpadded, in both places |

The ordering one is the sharper hazard: it sat inside the *"honour the data-layer gotchas"* list a DB-writing agent is told to follow, so it affected every read ordering rather than one number.

Table counts (13/22, now 14/23) and the SessionStart infra-gap section (closed by #231, #241) stay as historical record.

---

## 2. The guard, and why the existing one didn't fire

The AT band was wrong in **three** documents and took **three PRs** to find them all:

| File | Fixed by |
|---|---|
| `CLAUDE.md` | #266 |
| `web/.design-sync/CLAUDE.md` | #268 |
| `coach/CODE_TO_COACH_HANDOVER.md` | this PR |

A zone check already existed by #268. It didn't catch the third copy because it lived in `check-design-sync-entry.mjs`, which runs inside `ci-web.yml`'s lint job **gated on `web/**`** — so a change to `coach/` or the repo-root `CLAUDE.md` never triggered it. The guard was real; it just never ran on the files that needed it. **Widening the scan without widening the trigger would have repeated that**, so both halves move:

- **`web/scripts/check-zone-bands.mjs`** — scans every tracked `.md` via `git ls-files` and recomputes each published UT2/UT1/AT band against `derivePaceZones`, using the CP *each document itself quotes*, so a historical doc isn't judged against today's anchor. Removed from the design-sync guard, which keeps barrel + contrast + ownership.
- **`.github/workflows/zone-bands.yml`** — runs it on `pull_request` for `**/*.md`, `trainingConfig.js`, `pace.js` and the script. No `npm ci`; the script has no third-party imports.

### Two bugs the first cut had, both found by testing rather than reading

1. **It reported 1 band across 284 files when there are 3.** The separator class was `[^0-9]` — but the labels contain digits (`UT1`, `UT2`), so it couldn't step over them and only the grouped phrasing ever matched. `[\s\S]` fixes it. Had I trusted the passing run, the guard would have shipped blind to two of the three files it exists for.
2. Bands **wrap across lines** in `coach/` and carry markdown emphasis *inside* the numbers (`AT **164–185** W`), so the text is normalised before matching.

## How to test manually

```
npm run check:zones
# check:zones: 3 published zone band(s) across 284 tracked markdown files match derivePaceZones.
```

Verified by mutating each file in turn — every one is reported independently:

| Mutation | Result |
|---|---|
| `164–185` → `164–205` in `CLAUDE.md` | 1 failure |
| same in `web/.design-sync/CLAUDE.md` | 1 failure |
| same in `coach/CODE_TO_COACH_HANDOVER.md` | 1 failure |
| `UT2 113–144` → `100–144` | `gives UT2 as 100-144 W; derivePaceZones(205) yields 113-144` |
| clean tree | passes |

This PR triggers the new workflow itself, so it self-tests.

## Checklist

- [x] CI is green — note the `web/**` gates path-filter out for `coach/`-only diffs; the new `Zone bands` job is filtered on `**/*.md` precisely so it *doesn't*
- [x] Tests cover the change, or I've noted why they don't — the guard is verified by mutation, matching how `check-design-sync-entry.mjs` was already built; locally 753 tests, lint, format, `check:design-sync` and `check:zones` all pass

### DIFF
--- a/.github/workflows/zone-bands.yml
+++ b/.github/workflows/zone-bands.yml
@@ -1,0 +1,44 @@
+name: Zone bands
+
+# Rowing zone bands are derived from the live rowing_cp anchor by
+# derivePaceZones() — they are not constants. Any watt figure written into prose
+# is a snapshot, and the same wrong AT ceiling (205 W, the CP anchor restated as
+# the band's top) reached three separate documents before anything caught it:
+# CLAUDE.md (#266), web/.design-sync/CLAUDE.md (#268) and
+# coach/CODE_TO_COACH_HANDOVER.md (#275).
+#
+# A zone check already existed by #268 and still missed the third copy, because
+# it lived in the design-sync guard — which runs inside ci-web.yml's lint job,
+# gated on 'web/**'. A change to coach/ or the repo-root CLAUDE.md never
+# triggered it. This workflow exists so the guard fires on the documents it is
+# meant to protect, not only on the ones that happen to sit under web/.
+#
+# The script imports trainingConfig.js directly and has no third-party
+# dependencies, so this needs node and a checkout — no npm ci.
+
+on:
+  pull_request:
+    paths:
+      - '**/*.md'
+      - 'web/src/constants/trainingConfig.js'
+      - 'web/src/utils/pace.js'
+      - 'web/scripts/check-zone-bands.mjs'
+      - '.github/workflows/zone-bands.yml'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: zone-bands-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check:
+    name: Zone bands match derivePaceZones
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-node@v7
+        with:
+          node-version: 22
+      - run: node web/scripts/check-zone-bands.mjs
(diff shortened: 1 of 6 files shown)


skizzy1986:
Internal change — no release note. This PR corrects a documentation snapshot (`coach/CODE_TO_COACH_HANDOVER.md`) and adds a new CI guard (`check:zones` + the `Zone bands` workflow) that verifies published zone bands against `derivePaceZones()`. No application code, UI, or user-facing behavior changed.

---

PR #281, opened 2026-08-24
fix: captureError mistitles non-Error exceptions, and swallowed Supabase errors

### LINKED ISSUE #276: fix: captureError mistitles non-Error exceptions, and query-read failures may not reach Sentry
Two findings from a post-merge verification pass on #271, made while confirming source maps resolve correctly (they do — confirmed via [ERG-DASHBOARD-2](https://splitiq-29.sentry.io/issues/ERG-DASHBOARD-2), a real `useMutation` failure with a fully symbolicated stack trace). Verified against the code before filing; the second finding still needs root-causing, not just fixing.

## 1. Non-`Error` objects lose their title — confirmed

`web/src/utils/sentry.js:54-57`:

```js
export function captureError(error, context) {
  if (!error) return;
  Sentry.captureException(error, context ? { extra: context } : undefined);
}
```

No `instanceof Error` check. Supabase (postgrest-js) errors are plain objects — `{code, details, hint, message}` — not `Error` instances. Passed straight to `Sentry.captureException`, Sentry falls back to a generic **"Object captured as exception with keys: code, details, hint, message"** title instead of a distinct one. This is exactly what ERG-DASHBOARD-2 shows, and it affects **every** Supabase failure that reaches `captureError`, not just this one — they'll all collapse under the same generic title, defeating grouping and triage.

**Fix:** wrap non-`Error` inputs before capturing, e.g.:
```js
const err = error instanceof Error ? error : new Error(error?.message || String(error));
```

## 2. `useQuery` (read) failures may never reach Sentry — needs diagnosis first

During the same verification pass, blocking Supabase client-side and triggering both reads (`useQuery` hooks — auth/user, anchors, vitals, coach_messages, sessions) and a write (`useMutation` — Log a Strength Session) showed: the write reported to Sentry correctly (ERG-DASHBOARD-2); **no read ever did**, despite retrying and ultimately failing.

`handleQueryError` and `handleMutationError` are wired identically in `web/src/main.jsx` (`QueryCache`/`MutationCache`, both via `web/src/utils/queryErrorHandlers.js`), so the asymmetry isn't in the handler wiring itself. Two real, undistinguished possibilities:

- **Timing** — `retry: 2` (global default in `main.jsx`) means a blocked query takes a few seconds of backoff before settling into an error state; the observation window may have ended before that happened.
- **A second instance of the swallow pattern already found once** — `useSessionLog.js` bypassed react-query entirely and needed its own explicit `captureError` call (added in #271). It's plausible one or more of the other ~10 read hooks (`useSessions.js`, `useVitals.js`, `useAnchors.js`, `useTSSHistory.js`, `useErgSessions.js`, `useBenchmarkSessions.js`, …) catch internally and return fallback data instead of throwing, in which case react-query never sees a failure and `QueryCache.onError` correctly never fires — same class of bug, different location.

**Do not fix blind.** First reproduce deliberately (block Supabase, wait past the full retry/backoff window, confirm the query actually reaches `status: 'error'` in React Query devtools) to tell these apart, then fix whichever is true — either nothing (working as designed, false alarm from an early check) or add the missing `captureError` call in whichever hook(s) swallow.

## Acceptance criteria

- [ ] `captureError` wraps non-`Error` inputs in a real `Error`; test covers a plain-object input producing a distinct, meaningful message
- [ ] Root cause of the read-vs-write asymmetry is identified (timing vs. a swallowing hook) and stated in the PR
- [ ] If a swallowing hook is found, it reports through `captureError` the same way `useSessionLog.js` does
- [ ] If it's a false alarm (timing only), a test or comment records that so it isn't re-investigated later

## Out of scope

Sentry Test Analytics/Prevent — separately confirmed not activated on the account (Business Plan Trial, no Prevent line item, `/prevent/tests/` 404s). Not a code issue; an account-activation question for Scott.

skizzy1986:
Closes #276

## What changed and why

Two related Sentry reporting gaps found in a post-merge verification pass on #271.

**1. `captureError` mistitled non-`Error` exceptions.** `web/src/utils/sentry.js`'s `captureError()` passed non-`Error` values (postgrest-js's plain `{code, details, hint, message}` shape) straight to `Sentry.captureException`, so every Supabase failure that reached it was titled the generic "Object captured as exception with keys: ...". A `toError()` helper now wraps non-`Error` inputs in a real `Error` using the postgrest message when present, preserves `code`/`details`/`hint` under `extra.originalError`, and sets a Sentry `fingerprint` on the wrapped path only so distinct postgrest errors from the same call site don't over-group under Sentry's stacktrace-led default grouping.

**2. Root cause of the read-vs-write reporting asymmetry — identified, and it's not what the original hypothesis guessed.** The 9 standard `useQuery` read hooks (`useSessions`, `useVitals`, `useAnchors`, `useTSSHistory`, `useErgSessions`, `useBenchmarkSessions`, `useStrengthPRs`, `useVitalsSync`, plus the `coach_messages` query in `useCoach`) all `throw` their Supabase error correctly and are reported by the `QueryCache`/`MutationCache` sink wired in `main.jsx`. The asymmetry actually observed is **retry-backoff timing**: queries retry twice with backoff (≈3s) before `onError` fires, while mutations have no retry override and report near-instantly — a short manual observation window ends before a blocked read finishes retrying. That's recorded in a comment next to `main.jsx`'s `retry: 2` so it isn't re-investigated.

Two **real** swallow bugs were found instead, in code that talks to `supabase-js` directly rather than through react-query:
- `useOfflineQueue.js`'s `drainQueue` silently retried genuine insert failures without reporting them (duplicate-key `23505` conflicts correctly stay silent — that's expected "already delivered" behaviour, verified by a negative-assertion test).
- `useCoach.js`'s three `coach_messages` inserts discarded their `{data, error}` result entirely — postgrest-js resolves rather than rejects without `.throwOnError()`, which this codebase never calls, so these failures were never observed, not merely caught-and-swallowed. `clearHistory`'s delete had the same gap, plus no try/catch at all around a rejecting `getUser()`.

Both now report through `captureError`, matching the `useSessionLog.js` precedent (`{source, op}` context shape). Failed coach-message inserts stay report-only (no throw, no send-UX change — a follow-up issue can decide whether to surface this in the UI). `clearHistory` reports a delete failure, now surfaces it to the user via the existing `error` state (previously a failed clear was a silent no-op), and no longer clears the UI's message list when the server-side delete actually failed.

## How to test manually

Covered by the extended Vitest suite (see below) — no manual steps needed.

## Checklist

- [x] CI is green (lint, test, build) — verified locally: `npm test` (772/772 passing), `npm run lint` (clean), `npm run format:check` (clean), `npm run build` (clean), `npx vitest run --coverage` (per-file gates on `sentry.js` and `useCoach.js`, plus the global floor, all clear)
- [x] Tests cover the change — new/extended cases in `sentry.test.js`, `useOfflineQueue.test.js`, `useCoachWiring.test.js`
- [x] Visual change: none
- [x] Stays inside the property boundary — no colour/spacing/type values touched; this is error-handling logic only

---

_Ran through the `/orchestrate` pipeline (research → spec → build → test verification → review + validate), with human approval gates on the spec and pre-ship review. Both judges (Code Reviewer, Reality Checker) reviewed the diff; two should-fix findings from the Reality Checker were applied before this push: a `??`/`||` fingerprint bug that would have silently defeated the over-grouping fix on every network-level failure (postgrest emits `code: ''`, not absent), and a silent UX dead-end where a failed `clearHistory` gave the user no feedback at all._

---
_Generated by [Claude Code](https://claude.ai/code/session_01U72PeXfLHwoLhY543pFyxJ)_

### DIFF
--- a/web/src/hooks/__tests__/useCoachWiring.test.js
+++ b/web/src/hooks/__tests__/useCoachWiring.test.js
@@ -8,7 +8,18 @@ import React from 'react';
 // function, and vi.mock is file-scoped and hoisted, so a test that needs the
 // real hook machinery cannot live alongside it.
 
-const insertMock = vi.fn(() => Promise.resolve({ error: null }));
+const insertMock = vi.fn((row) =>
+  Promise.resolve({ error: insertErrorFor(row) })
+);
+const deleteEqMock = vi.fn(() => Promise.resolve({ error: deleteError }));
+const getUserMock = vi.fn(() => getUserResult());
+
+// Same mutable-module-state idiom as sessionRows/vitalsState below: the mock
+// factory is hoisted, so per-test overrides have to reach it through bindings.
+// insertErrorFor takes the row so a test can fail one role's insert only.
+let insertErrorFor = () => null;
+let deleteError = null;
+let getUserResult = () => Promise.resolve({ data: { user: { id: 'u1' } } });
 
 vi.mock('../../supabaseClient.js', () => ({
   supabase: {
@@ -19,15 +30,19 @@ vi.mock('../../supabaseClient.js', () => ({
           limit: () => Promise.resolve({ data: [], error: null }),
         }),
       }),
-      delete: () => ({ eq: () => Promise.resolve({ error: null }) }),
+      delete: () => ({ eq: (...args) => deleteEqMock(...args) }),
     }),
     auth: {
       getSession: () =>
         Promise.resolve({ data: { session: { access_token: 'test-token' } } }),
-      getUser: () => Promise.resolve({ data: { user: { id: 'u1' } } }),
+      getUser: (...args) => getUserMock(...args),
     },
   },
 }));
+
+vi.mock('../../utils/sentry.js', () => ({ captureError: vi.fn() }));
+
+import { captureError } from '../../utils/sentry.js';
 
 let sessionRows = [];
 let vitalsState = {
@@ -101,7 +116,13 @@ beforeEach(() => {
   sessionRows = [];
   tssRows = [];
   vitalsState = { latest: null, readinessScore: 0, readinessLabel: 'FATIGUED' };
+  insertErrorFor = () => null;
+  deleteError = null;
+  getUserResult = () => Promise.resolve({ data: { user: { id: 'u1' } } });
   insertMock.mockClear();
+  deleteEqMock.mockClear();
+  getUserMock.mockClear();
+  captureError.mockClear();
   localStorage.clear();
 });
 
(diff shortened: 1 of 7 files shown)


skizzy1986:
**Release note draft**

**Fixed** — Clearing your Coach chat history now tells you if it fails instead of failing silently: if the server-side delete errors, you'll see an error message and the chat list stays put rather than getting wiped from the screen while the messages are still there in the background.

The rest of this PR is backend error-reporting plumbing (Sentry now categorizes Supabase errors correctly and reports a couple of previously-silent failure paths) with no other visible effect on the app.

---

PR #277, opened 2026-08-24
refactor(theme): rename the colour-named THEME keys to roles — values unchanged

skizzy1986:
Closes #248

## What changed and why

The 10 colour-named `THEME` keys become role names so the single source of colour truth names meaning instead of pigment — the first of four palette steps, unblocking the `var(--color-*)` seam (#250/#251). **No hex value moves.** All 23 keys survive; the key count and insertion order are unchanged.

| Old | New | Hex | App call sites |
|---|---|---|---|
| `cyan` | `accent` | `#00d4ff` | 11 |
| `green` | `positive` | `#34d399` | 22 |
| `gold` | `caution` | `#ffd700` | 12 |
| `orange` | `warning` | `#ff6b35` | 3 |
| `red` | `critical` | `#ff2d55` | 10 |
| `purple` | `accentAlt` | `#a78bfa` | 0 |
| `pink` | `accentAlt2` | `#f472b6` | 0 |
| `teal` | `positiveAlt` | `#2dd4bf` | 0 |
| `grey` | `neutralAccent` | `#888888` | 4 |
| `white` | `textStrong` | `#ffffff` | 3 |

The 13 structural keys (`bg`, `surface`, `raised`, `field`, `border`, `text`, `muted`, `surfaceAlt`, `neutral`, `divider`, `textSubtle`, `textFaint`, `textDim`) were already role-named and are untouched.

### The three names the issue left open

- **`white` → `textStrong`** — both sites are maximum-emphasis foreground one step above `text`: the bold CTL/ATL numerals in `LoadTooltip.jsx` and the `fontWeight: 700` session title in `LogEntry.jsx`. It completes the existing `text → textSubtle → textFaint → textDim` ramp upward. A key named `white` holding a dark hex after the light flip would be actively misleading.
- **`grey` → `neutralAccent`** — the unifying role across all four sites is "no status applies". `LogEntry.jsx:7` is `C[entry.type] || THEME.grey`, the accent-slot fallback, used as border, background tint *and* text — so a text-only name would be wrong. Deliberately distinct from the structural `neutral` (`#3a3a4a`), which is a surface tone; this is a foreground one.
- **`teal` → `positiveAlt`** — `HANDOFF.md` diagnoses it as a duplicate role of `positive`, and `#2dd4bf` is a green-cyan genuinely in that family. Naming it this way preserves the diagnosis so a later PR can fold it knowingly.

### Two deliberate divergences from HANDOFF.md

[...]

Issue #250, opened 2026-08-23
refactor: make THEME values var(--color-*) and retire the 8-digit-alpha convention


skizzy1986:
Second palette step, after the role rename. `web/.design-sync/HANDOFF.md` §1.

## The inversion the handoff does not mention

`web/src/utils/themeCss.js` `cssVars(THEME)` emits `--color-<key>: <value>` — injected at `main.jsx:22` and used by `.design-sync/gen-css.mjs` to build `base.css`. Once `THEME.accent` is `'var(--color-accent)'`, that emits:

```css
--color-accent: var(--color-accent);   /* self-referential, resolves to nothing */
```

Today THEME is the source and the CSS is generated. **After the seam the CSS is the source and THEME is a pointer table.** So `theme.js` splits in two:

- a **values module** — the hexes, consumed by `cssVars()` and `gen-css.mjs`
- **`THEME`** — every value becomes the string `var(--color-<key>)`

Plus a real stylesheet, `web/src/styles/theme.css`: `:root, [data-theme='light']` carrying all 20 values. Light is also `:root`, so `data-theme` is a no-op. **No dark block ships.** `main.jsx` imports it instead of injecting generated vars.

## Tests that break here

All six derived-value files, because jsdom does not resolve custom properties — rewrite them to assert the **token string**, which is simpler and more honest than trying to resolve a colour:

`formatting.test.js` (23 hex lines, `workoutAccent()`), `analysis.test.js`, `recoveryAnalytics.test.js`, `OverviewView.test.jsx`, `BenchmarkBadge.test.jsx`, and `theme.test.js` (its `/^#[0-9a-f]{6}$/` regex moves to the values module; `THEME` gets a new assertion that every value is exactly `var(--color-<key>)` and the key sets match).

Also re-point `e2e/smoke.spec.js:108` at the values module rather than the literal `rgb(0, 212, 255)`, so the light flip need not touch e2e.

## The alpha collision

`#183`'s idiom is `` `${THEME.token}NN` `` 8-digit hex. `var(--color-x)80` is **invalid CSS and fails silently** — the background just vanishes. Four sites:

- `web/src/components/LogSessionForm.jsx:144`
- `web/src/components/WorkoutItem.jsx:17`
- `web/src/components/LogEntry.jsx:78` and `:112`

[...]

Issue #248, opened 2026-08-23
refactor: rename the colour-named THEME keys to roles — values unchanged


skizzy1986:
First of the four palette steps. See `DESIGN.md` and `web/.design-sync/HANDOFF.md` §1.

The rename ships **separately from the seam, with no value changed**, because that is the only version of this change that seven colour-locking test files let through cleanly.

## The rename

| Old | New |
|---|---|
| `cyan` | `accent` |
| `green` | `positive` |
| `gold` | `caution` |
| `orange` | `warning` |
| `red` | `critical` |
| `purple` | `accentAlt` |
| `pink` | `accentAlt2` |
| `teal` | fold into `positive` |
| `grey`, `white` | decide explicitly — delete, or keep |

Structural keys (`bg`, `surface`, `raised`, `field`, `border`, `text`, `muted`, `surfaceAlt`, `neutral`, `divider`, the `text*` ramp) are already role-named and stay.

## Scope

- `web/src/constants/theme.js` and `theme.test.js` (`EXPECTED_KEYS` + the canonical-values block — same hexes, new names)
- ~71 colour-named `THEME.*` call sites in `web/src/`
- **The four-way reconciliation**: all 15 colour-named references in `web/.design-sync/docs/*.md`, `previews/*.tsx` and `config.json`'s `dtsPropsFor` defaults. These are currently correct against the dark code; they move in lockstep here.
- Re-run `node .design-sync/gen-css.mjs`

## Not touched

The six derived-value test files (`formatting.test.js`, `analysis.test.js`, `recoveryAnalytics.test.js`, `OverviewView.test.jsx`, `BenchmarkBadge.test.jsx`) and `e2e/smoke.spec.js`. They assert *values*, which do not change. That is the point of doing the rename first.

## Acceptance

- [ ] `git diff -- web/src/constants/theme.js | grep -E '^[+-].*#[0-9a-f]{6}'` shows every hex on both a `+` and a `-` line — no value moved
- [ ] Screenshot baselines byte-identical (needs the baseline harness first)
- [ ] The six derived-value test files pass **unmodified**
- [ ] `node .design-sync/gen-css.mjs && git diff --exit-code web/.design-sync/base.css` clean
- [ ] No colour-named key remains anywhere in `web/.design-sync/`

Blocked by the baseline harness. Blocks the `var(--color-*)` seam.

Property boundary: this touches colour identifiers only — no sizes. Related to #183.

skizzy1986:
[...]

PR #282, opened 2026-08-24
docs(design): bring down two design drops, and link repo-owned docs instead of uploading them

skizzy1986:
Refs #258, #269, #248, #277.

Two design-project exports imported, the Code → Design bulletin refreshed, and one mechanism finding that changes how the two sides exchange documents.

---

## 1. Design can read the repo — so repo-owned docs are linked, not uploaded

This was never written down. Every document states the constraint in **one direction only** — *"a code session cannot read this project"* — and the reverse was assumed rather than checked, including by the briefing I wrote.

An uploaded copy is a snapshot, and snapshots go stale. That is the failure this workstream keeps paying for: the bundle in the project still serves pace bands frozen at a CP of 190 *because* it holds a copy. A link cannot go stale.

So **the "upload two documents to the project root" half of #258 dissolves** rather than gets satisfied.

**The test that confirmed it also found the limit, which is worth more.** Asked to read `CODE-TO-DESIGN.md` and quote its headline, a design session returned one dated four PRs earlier — correctly, because the refresh was on an unmerged branch. **Linking resolves to `main`.** Unmerged work is invisible to design exactly as it is to a fresh code session.

`NOTES.md` now records the asymmetry with both halves stated, since assuming symmetry is what produced the wrong briefing:

| Direction | |
|---|---|
| Code → project | **Blind.** Design-owned files travel by hand-off, land by a code commit |
| Design → repo | **Readable.** Read-only, at `main` |

## 2. Two design drops imported

| | |
|---|---|
| **`splitiq-light-tokens.css`** | The light role-named palette. Now maps onto the code's existing key names (`surface-alt`, `raised`, `field`, the full text ramp) and splits `text-subtle` `#43485a` from `muted` `#4a4f63`, stating plainly that `text-faint` clears 3:1 on a card only and `text-dim` clears nothing anywhere |
| **`splitiq-load.js`** | A shared CTL/ATL/TSB + readiness model the artboards read, so two screens cannot state different numbers for the same day — "computed, never typed" made concrete |
| **`designs/desktop-overview.html`** | First desktop artboard |
| `HANDOFF.md`, `PROJECT-CONTEXT.md`, `conventions.md` | Revised on the design side |

**Verified rather than trusted: 34 ratio claims in the token file, 34 checked, 0 mismatches.** My first sweep matched only 3 — the separator had double spaces — so the check now reports its own coverage. "All 3 verify" out of 34 would have been a vacuous pass.

**Everything raised from this side came back actioned:** the `#c3cade`/`#bcc5dd` ground inconsistency is gone, artboard thumbnails paint the real ground, `HANDOFF` §1's *"Dark: dropped"* now reads *"retained as a second theme"* with an explicit supersession note, and `PROJECT-CONTEXT` records the light-target vs token-names distinction my briefing had collapsed.

## 3. The ownership guard earned itself, then found a hole in itself

`check:design-sync` **refused the first import** until the two new files were classified — exactly the decision it exists to force.

Then the pre-commit hook rewrote `splitiq-load.js` from 87 lines to 232. `lint-staged` matches `*.{js,jsx}` with `matchBase`, so it reached a design-owned mirror; the manifest declared the direction but nothing stopped the toolchain acting against it. Fixed at the tools (`.prettierignore` + eslint `ignores`) rather than by narrowing the glob, which would have changed `matchBase` semantics and quietly dropped top-level JS from the hook. Restored byte-identical and verified by re-running the hook against a checksum.

## 4. Bulletin refreshed — the headline inverts

The previous edition said *"do not re-sync yet"*, gated on the rename. **#248 closed** (merged as #277), so both its prerequisites are satisfied and **the re-sync is unblocked**. Its entire "what Code needs from Design" table is closed out: all nine missing token values supplied, ground settled, artboards and `ISSUES-load-states.md` committed.

Corrected by age: it listed #249 as pending (closed by #267) and named Archivo for `cfg.extraFonts` — superseded by IBM Plex per `STATE_OF_PLAY` §4.1.

**One open disagreement carried forward:** #277 kept `teal` rather than folding it — agreeing with the light palette on substance — but named it **`positiveAlt`** where the palette says **`cycling`**. That reintroduces by name the conflation both sides rejected. Cheap to converge before the seam (#250) wires the token up.

## How to test manually

```
npm run check:design-sync   # 45 files classified, contrast claims recompute
npm run check:zones         # 3 bands across 284 tracked markdown files
```

Mirror integrity, the bug this PR fixes:

```
md5sum web/.design-sync/splitiq-load.js   # before and after a commit — unchanged
```

## Checklist

- [x] Tests cover the change, or I've noted why they don't — documentation and mirrors; the guards are verified by mutation, and locally 772 tests, lint, format, `check:design-sync` and `check:zones` all pass

---


skizzy1986:
Internal change — no release note. This PR imports design-project artifacts (tokens, load-model mirror, an artboard) and refreshes design-sync docs/tooling; no runtime code changed and nothing in the shipped app is user-visible from this diff.

---

PR #284, opened 2026-08-24
feat(mobile): animated SplitIQ splash for the cold-start window

### LINKED ISSUE #283: Animated SplitIQ splash for the mobile cold-start window
Opening the app on mobile currently shows a bare `Loading…` text node on a black background while `supabase.auth.getSession()` resolves and the first tab fetches. It reads as an unfinished app rather than a product.

A splash design was produced in Claude Design (four concepts; 2a — "stroke-curve mark, refined" — chosen): an animated cyan stroke-curve mark drawn inside a rounded tile with a pulsing radial glow and expanding halo, the `SplitIQ` wordmark rising in, an `ERG · STRENGTH · BIKE` subtitle whose letter-spacing tightens, an indeterminate progress track, and a `SYNCING TRAINING LOG` caption.

## Acceptance criteria

1. On mobile, the splash replaces the bare `Loading…` while `session === undefined`. Never both.
2. It also covers first-data-load — clears when the active tab's initial data has settled, **or** at a 5000 ms hard ceiling, whichever comes first.
3. Minimum display 700 ms from mount, so a warm start does not flash.
4. The animation loops for as long as it is mounted; it does not play once and freeze.
5. Caption is the static literal `SYNCING TRAINING LOG` — no boot-phase narration.
6. If `getSession()` rejects, the splash exits immediately (bypassing the floor) and hands off to Login.
7. A first-fetch failure is not splash-blocking; the tab shows its own error state.
8. `prefers-reduced-motion: reduce` renders the static end state. Timing unchanged — only motion is removed.
9. Shows on every cold start. Never re-appears on tab switch, resize, resume, or background refetch.
10. The mockup's fake status bar (`5:41`, battery glyph) is not shipped.
11. All of the above assertable in Vitest/RTL.

A clean logged-out cold start also shows the splash, clearing at the 700 ms floor rather than riding to the ceiling.

## Out of scope

Desktop splash; dynamic caption text; `@capacitor/splash-screen` or bridging the native Android launch screen; once-per-day suppression; background/foreground resume mid-splash.

skizzy1986:
Closes #283

## What changed and why

Replaces the bare `Loading…` text on the mobile cold-start path with the animated SplitIQ splash from the approved design (concept 2a), gated on a 700 ms floor / 5000 ms ceiling so it neither flashes on a warm start nor traps you on a slow one.

### How it is put together

All the decision-making lives in coverage-included modules; `main.jsx` (which is coverage-excluded) keeps only composition:

| File | Role |
|---|---|
| `utils/splashCss.js` | Pure `splashCss(theme)` → CSS string. Nine namespaced `siq-splash-*` keyframes. |
| `hooks/useSplashGate.js` | The floor/ceiling/auth-fail state machine. Booleans in, boolean out — no react-query, no DOM. |
| `hooks/useInitialDataReady.js` | Self-unsubscribing react-query cache subscription; latches once and tears down. |
| `hooks/usePrefersReducedMotion.js` | Guarded `matchMedia`. |
| `components/mobile/SplashScreen.jsx` | The visual. No props. |

Three decisions worth flagging for review:

- **It overlays `MobileApp` rather than replacing it.** The tabs' fetches only start once `MobileApp` mounts, so it has to be mounted and fetching underneath — otherwise nothing would ever fetch and the splash would ride to the ceiling every time.
- **Readiness is a cache subscription, not `useIsFetching()`.** `useIsFetching` re-renders its host on every fetch start/stop for the process lifetime; hosted in `AuthGate` that means re-rendering the whole tree on every background refetch, forever. The subscription unsubscribes permanently once it latches. **Zero edits to `MobileApp` or any tab.**
- **`AuthGate` returns one flat shape across all three branches.** Returning a differently-shaped fragment per branch remounted the splash the moment the session resolved — React reconciles fragment children by index, so index 0 went `SplashScreen` → `SignOutButton`, tearing it down and replaying every entrance animation mid-flight. Measured 2 mounts before, 1 after.

## How to test manually

Open the app on a mobile viewport (or throttle the network). The splash should cover the boot, clear no sooner than 700 ms and no later than 5 s, and never re-appear when switching tabs. With OS-level reduced motion on, it should render fully-formed and still.

## Screenshots

Captured in Chromium at a Pixel 7 viewport against the production build — mid-loop, and the `prefers-reduced-motion` static end state.

| Animated (mid-loop) | Reduced motion (static end state) |
|---|---|
| Stroke drawn, head dot tracking the curve, glow pulsing, progress track scanning | Stroke complete, head dot and halo hidden, subtitle at rest, track empty |

## Notes for the reviewer

- **One new colour token.** `surfaceDeep: '#12121f'` — the tile gradient's second stop, which had no existing token (it sits between `bg` and `surfaceAlt` on the same ramp). Purely additive: renames nothing, no existing component reads it. `.design-sync/base.css` regenerated via `gen-css.mjs`, not hand-edited.
- **Two permitted colour literals** in `splashCss.js`: `rgba(0, 0, 0, .55)` and `rgba(255, 255, 255, .05)` on the tile's box-shadow. Neutral shadow/highlight alphas with no `THEME` token; inventing one would be worse. Every other colour, including the translucent cyans, derives from `THEME` through a module-private `alpha()` helper — proven by substitution tests rather than a hex grep.
- **`e2e/visual.spec.js` gains a one-line wait.** `toBeVisible()` ignores occlusion, so without it the three `mobile shell` baselines would capture the splash. No baselines should need regenerating — the `visual` project already sets `reducedMotion: 'reduce'` and `animations: 'disabled'`, and the splash unmounts atomically with no exit transition.
- **The head dot deliberately has no `cx`/`cy`.** `offset-path` composes as a pure translation, so setting them would double the offset and park the dot outside the viewBox — verified in Chromium (`cx=30 cy=88` → renders at `(60, 176)`). An `@supports not (offset-path: …)` block hides it on engines without motion-path support instead.
- **`setDismissed` during render** in `useSplashGate` is React's documented "adjust state during render" pattern, not a workaround — the `useEffect` form is a hard error under this repo's `react-hooks/set-state-in-effect`.
- **Known gap, pre-existing:** the five lines of prop wiring in `AuthGate` are untested, because `main.jsx` is coverage-excluded and not importable in jsdom (importing it runs `initSentry()` and `createRoot().render()` at module scope). An inverted flag there would show a bounded 5-second splash, and no test would catch it. Worth a follow-up to extract `AuthGate` into its own file and off the exclude list — deliberately not done here, as it would widen this PR into a refactor.

## Checklist

- [x] Tests cover the change, or I've noted why they don't — 41 new tests; the `main.jsx` wiring gap is noted above
- [x] Visual change: screenshots attached above
- [x] Stays inside the property boundary: colour hexes = #183 · box metrics = S6 · type = S-1 (see DESIGN.md)

On the property boundary: the boundary governs edits to existing styled lines, so three concurrent workstreams do not collide in the same lines. Every style here is on lines that did not exist before. Every colour reads from `THEME`, so #183's rename moves the splash mechanically like any other component; the one `theme.js` edit is additive; spacing and type live entirely inside the two new files for S6 and S-1 to retro-fit; and no existing styled line in any other file is touched.

https://claude.ai/code/session_01K6Px2NPhrf3bk21X4mYnYU

---
_Generated by [Claude Code](https://claude.ai/code/session_01K6Px2NPhrf3bk21X4mYnYU)_

### DIFF
--- a/web/.design-sync/base.css
+++ b/web/.design-sync/base.css
@@ -20,6 +20,7 @@
   --color-accentAlt2: #f472b6;
   --color-positiveAlt: #2dd4bf;
   --color-surfaceAlt: #1e1e30;
+  --color-surfaceDeep: #12121f;
   --color-neutral: #3a3a4a;
   --color-textSubtle: #aaaacc;
   --color-textFaint: #6c6c88;
--- a/web/e2e/visual.spec.js
+++ b/web/e2e/visual.spec.js
@@ -53,6 +53,11 @@ async function openAuthed(page, context) {
   await installAppFixtures(context, { now: FIXED_TIME });
   await page.clock.setFixedTime(FIXED_TIME);
   await page.goto('/');
+  // The mobile splash overlays the tab tree, and toBeVisible() ignores
+  // occlusion, so every mobile baseline would capture the splash instead.
+  // The timeout must beat the default 5000ms — that is exactly the splash
+  // ceiling, so a default wait races it.
+  await expect(page.locator('.siq-splash')).toHaveCount(0, { timeout: 10_000 });
 }
 
 // Tab switching is state-only (App.jsx and MobileApp.jsx both hold `view` in
--- a/web/src/components/mobile/SplashScreen.jsx
+++ b/web/src/components/mobile/SplashScreen.jsx
@@ -1,0 +1,65 @@
+import { THEME } from '../../constants/theme.js';
+import { splashCss } from '../../utils/splashCss.js';
+import { usePrefersReducedMotion } from '../../hooks/usePrefersReducedMotion.js';
+
+const CSS = splashCss(THEME);
+
+const MARK_PATH = 'M30 88 C 44 88, 46 40, 62 40 C 78 40, 80 88, 94 88';
+
+export default function SplashScreen() {
+  const reduced = usePrefersReducedMotion();
+  return (
+    <div
+      className={reduced ? 'siq-splash siq-splash--still' : 'siq-splash'}
+      style={{
+        position: 'fixed',
+        inset: 0,
+        zIndex: 2000,
+        background: THEME.bg,
+        display: 'flex',
+        flexDirection: 'column',
+        alignItems: 'center',
+        justifyContent: 'center',
+        gap: 36,
+        fontFamily: 'ui-sans-serif, system-ui, -apple-system, sans-serif',
+      }}
+    >
+      <style>{CSS}</style>
+      <div className="siq-splash__glow" />
+      <div className="siq-splash__mark">
+        <div className="siq-splash__halo" />
+        <div className="siq-splash__tile" />
+        <svg
+          width="124"
+          height="124"
+          viewBox="0 0 124 124"
+          style={{ position: 'relative', overflow: 'visible' }}
+        >
+          <line className="siq-splash__base" x1="30" y1="88" x2="94" y2="88" />
+          <path className="siq-splash__stroke" d={MARK_PATH} />
+          {/* Left at the SVG origin deliberately: offset-path translates the
+              element to the path point, so a circle authored at cx/cy would land
+              at twice the offset. Engines without motion path hide it instead. */}
+          <circle className="siq-splash__head" r="4" />
+        </svg>
+      </div>
+      <div
+        style={{
+          display: 'flex',
+          flexDirection: 'column',
+          alignItems: 'center',
+          gap: 12,
+        }}
+      >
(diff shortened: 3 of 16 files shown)


skizzy1986:
**Release note (New)**

Added an animated splash screen for the mobile app's cold start. Instead of a plain "Loading…" message, you'll now see the SplitIQ logo animate in while your training data loads — it shows for at least 0.7s (so it doesn't flash on a fast load) and at most 5s (so a slow connection never leaves you stuck on it), and respects your device's reduced-motion setting.

---

PR #286, opened 2026-08-24
refactor(boot): extract AuthGate from main.jsx so the boot wiring is testable

### LINKED ISSUE #285: Extract AuthGate from main.jsx so the boot wiring is testable
Follow-up from #284 (mobile splash). Both review stages flagged the same gap independently, and it was deliberately left out of that PR to avoid widening a feature change into a refactor.

## The problem

`web/src/main.jsx` is in the coverage `exclude` list, documented as "auth/bootstrap entry point, not unit-testable in jsdom". That is accurate as things stand: the module runs `initSentry()` (line 30), appends the theme `<style>` to `document.head` (line 35) and calls `ReactDOM.createRoot(...).render(...)` (line 346) at module scope, and `AuthGate` is not exported. Importing it in a test executes all of that.

The consequence is that `AuthGate` (line 271) contains the only untested decision surface on the boot path. #284 raised the stakes by adding a five-line boolean handoff there:

```js
const showSplash = useSplashGate({
  enabled: isMobile,
  authResolved: session !== undefined,
  authFailed,
  dataReady,
  dataExpected: !!session,
});
```

Every hook it feeds is at 100% coverage. The wiring that feeds them is at zero.

## Silent regressions nothing would currently catch

1. **`dataExpected: !!session` dropped or inverted.** A clean logged-out cold start never latches readiness (no query ever runs), so the splash rides to the 5000 ms ceiling before Login appears — precisely the failure the `dataExpected` flag was added to prevent.
2. **`authResolved: session !== undefined` written as `!!session`.** Same 5-second dead splash on every signed-out launch, different cause.
3. **The `.catch` handler or `setAuthFailed(true)` removed** (line 291-292). An auth outage parks on the splash for the full ceiling instead of exiting immediately to Login.
4. **`enabled: isMobile` hardcoded true.** Splash on desktop.
5. **The `session === undefined` ternary reverted to rendering both children.** The splash and the bare `Loading…` render together.

The e2e guard in `web/e2e/visual.spec.js` (`toHaveCount(0, { timeout: 10_000 })`) is a weak backstop for 1–3 on the authed path only: it asserts the splash *goes away*, never that it *appeared*, so it would pass identically if the splash never rendered at all. Its baselines are also CI-container-only.

## Proposed change

A strangler-fig extraction of exactly the kind `/refactor` exists for:

1. Move `AuthGate` to `web/src/AuthGate.jsx` and export it. Leave `main.jsx` as pure bootstrap — `initSentry()`, the theme style tag, the QueryClient, and `createRoot().render(<AuthGate />)`.
2. Drop `src/AuthGate.jsx` from the coverage `exclude` list in `web/vite.config.js` and add a per-file threshold at the usual 80/80/70. `main.jsx` stays excluded.
3. Land tests naming failures 1–5 above explicitly, mocking `supabaseClient` and the splash hooks the way `views/mobile/__tests__/` already mocks hooks.
4. Optionally add the cheap black-box assertion to `openAuthed()` — that `.siq-splash` **is** visible immediately after `page.goto('/')`, before the existing wait for it to clear.

## Notes

- This is pre-existing debt, not something #284 created — `AuthGate`'s auth branching was already untested. #284 inherits the gap and makes it more load-bearing.
- Worst case from any of the five failures is a bounded 5-second splash: no data loss, no broken app, no RLS bypass. Worth fixing, not urgent.
- Related latent issue found while working on #284: `web/src/hooks/useIsMobile.js` calls `window.matchMedia` unguarded in its `useState` initialiser, which throws in jsdom (there is no `matchMedia` there at all, and `test-setup.js` stubs only `ResizeObserver`). That is why the hook has no tests. **Anything that guards it must also update `e2e/visual.spec.js`** — the `error fallback` baseline deliberately forces a render error by making `window.matchMedia` throw, and relies on `useIsMobile` being the call site.

skizzy1986:
Closes #285

## What changed and why

`AuthGate` moves out of `main.jsx` into `web/src/AuthGate.jsx` and gains tests, so the five-line splash/auth handoff that was the only untested decision surface on the boot path is now gated by CI.

The extraction itself is a **pure move** — 283 lines byte-identical, the sole delta being `export default` added to `AuthGate`. All the new value is in the tests. `main.jsx` drops from 356 lines to 64 of pure bootstrap: `initSentry()`, the theme `<style>` injection, the Capacitor branch, the `queryClient` construction, and the root render.

`Login`, `SignOutButton`, `InstallButton` and the private `C` token map travel with it. `InstallButton` in particular could not stay behind — it is one of the three children of the flat fragment whose child index the splash entrance animation depends on.

### Two test files, two strategies

Both are needed, and neither is redundant:

- **`AuthGate.wiring.test.jsx`** (W1–W6) mocks `useSplashGate` as a spy and asserts the *complete literal* argument object, so it fails on a wrong value **and** on a dropped key.
- **`AuthGate.test.jsx`** (B1–B13) runs the real gate under fake timers and asserts what the DOM does at the 700 ms floor and 5000 ms ceiling.

A logged-out cold start looks identical whether `dataExpected` is inverted or `authResolved` was written as `!!session` — behaviour alone cannot tell them apart. Conversely a spy alone proves nothing about the floors.

### Mutation evidence

Every regression was physically reintroduced and confirmed to fail a named assertion:

| Mutation | Fails |
|---|---|
| `dataExpected` key dropped | W1, W2, W3, W4, W6 · B2 |
| `dataExpected: !session` (inverted) | W1–W4, W6 · B1, B2, B9, B12 |
| `authResolved: !!session` | W2, W4, W6 (W1 correctly passes) · B1, B9, B12 |
| `.catch` body emptied | W4 · B3 |
| `enabled: true` hardcoded | W5 · B4, B5 |
| Ternary reverted, both children render | B5, B6 |
| Outer fragment special-cased per branch | B6 |
| Outer fragment special-cased on `session === null` only | B1 |

The last two are complementary: `!session` is also true while the session is still `undefined`, so that variant leaves B1's two states sharing a shape and only B6 catches it. Isolating `session === null` inverts which one bites. Together they evidence the child-index invariant on all three render branches.

### Two spec corrections found while building

**`getTimerCount()` cannot be 0 immediately after the floor fires.** `useSplashGate`'s timer effect depends on `[active]` (`enabled && !authFailed`), and dismissal happens through a render-time latch that leaves `active` untouched — so the cleanup never runs and the ceiling stays armed. B1 and B7 assert `1` there and `0` only past the ceiling. B3 (auth failure) and B4 (desktop) keep `0`, where `active` genuinely does go false or was never true.

**The theme-token banner comment** moved with its block rather than being orphaned above `ReactDOM.createRoot`.

### Coverage

`AuthGate.jsx` lands at **100% lines / 100% functions / 96.66% branches**, against a new per-file threshold of 80/80/70. `exclude` is untouched — `main.jsx` stays excluded.

Global coverage **rose** to 87.21 / 81.78 / 79.46 (from 84.31 / 79.33 / 76.31): those ~285 lines were previously invisible inside excluded `main.jsx` and now enter the denominator fully covered. No threshold was lowered.

## How to test manually

Covered by CI. The one thing CI checks that could not be run locally is the Playwright visual suite — no browser cache in the agent container — so `openAuthed()` passing rests on the move being byte-identical and `visual.spec.js` being absent from the diff.

## Checklist

- [x] CI is green (lint, test, build) — 832 tests / 67 files, lint 0, format 0, build 0 locally; CI to confirm
- [x] Visual change: none — pure move, no rendered output changed
- [x] Stays inside the property boundary: no colour, box-metric or type value was touched. The pre-existing raw hex `'#08080dcc'` (`AuthGate.jsx:174`) moved verbatim; normalising it here would have broken the pure-move contract and belongs to #183.

## Follow-ups deliberately not included

- `useIsMobile` reads `window.matchMedia` unguarded — guard for **absence**, not for *throwing*, or it breaks the e2e error-fallback baseline that depends on a throwing `matchMedia`
- The raw hex above, and stale `main.jsx` references in `e2e/visual.spec.js:136,167` and `e2e/fixtures.js:4`
- The dead ceiling timer costs one full re-render of `MobileApp`/`App` five seconds into every mobile boot — pre-existing in `useSplashGate`
- Global threshold ratchet: headroom is now ~7 points

---
_Generated by [Claude Code](https://claude.ai/code/session_0114NeA5UtKh3HFk3UVzemAn)_


skizzy1986:
Internal change — no release note. This is a pure refactor (moves `AuthGate` out of `main.jsx` into its own tested file; byte-identical logic) with no user-visible behavior change.

---

PR #289, opened 2026-08-25
ci: have Update Visual Baselines commit the PNGs to the branch

### LINKED ISSUE #288: ci: have Update Visual Baselines commit the PNGs instead of uploading them
### What needs doing

`e2e-web.yml`'s **Update Visual Baselines** job regenerates the nine screenshot baselines inside the pinned Playwright container and then uploads them as an artifact. Committing them is a manual step: download the zip, unzip into `web/e2e/visual.spec.js-snapshots/`, commit, push. The workflow header documents that explicitly — *"run it from the Actions tab when a design change is meant to move pixels, download the artifact, commit the PNGs."*

The manual half is the weak link, for two reasons.

**It cannot be automated by anything that isn't a browser.** Baselines have to be produced in `mcr.microsoft.com/playwright:v<pinned>-noble`, because glyph rasterisation differs between images and a baseline made anywhere else fails forever. So the generation genuinely has to happen in CI — but the artifact then can only be retrieved by a human with a browser session. An agent session cannot: the GitHub API is reachable only through MCP tools, which expose no artifact-download call, and there is no Docker daemon to reproduce the image locally. This came up for real on #287, where the flip moved every pixel on all nine baselines and the run had to stop at "here is a link, please download it".

**It is the step most likely to be skipped.** A design PR is already green on its required checks by then (Visual is not required), so the incentive to finish the loop is weak, and a stale baseline set quietly stops testing anything.

### Area

infra (CI / workflows), testing

### Proposal

When dispatched with `update_baselines` on a branch, have the job commit the regenerated PNGs to that branch and push, instead of (or as well as) uploading the artifact.

Worth deciding deliberately rather than just doing, because it changes the pipeline's security posture: the job needs `contents: write` and a token that can push to the branch. Points to settle:

- **Scope the permission.** `permissions: contents: write` on that job only, not workflow-wide. Everything else in `e2e-web.yml` stays read-only.
- **Refuse to push to `main`.** The job is `workflow_dispatch`-only today, but it should still guard `github.ref != 'refs/heads/main'` so a baseline commit can never land unreviewed on the default branch.
- **Keep the artifact.** Uploading as well as committing costs nothing and preserves the current escape hatch if the push is blocked or the branch is protected.
- **Make the commit legible.** A fixed message naming the run (`test: regenerate visual baselines (run <id>)`) so `git log` shows where the PNGs came from and against which head SHA.
- **Consider whether it should push at all vs. open a PR.** Pushing straight to the feature branch is simpler; a bot PR is more reviewable but adds a second thing to merge. Pushing is probably right given the branch is already under review.

One thing this does *not* fix, and should not pretend to: committing baselines automatically still accepts whatever the screenshots now show. The human judgement — *is this what the app should look like?* — belongs on the PR, not in the job. This only removes the clerical step of moving files.

### Definition of done

- [ ] The dispatch job commits and pushes the regenerated baselines to the dispatched branch
- [ ] `contents: write` is scoped to that job alone; the rest of the workflow stays read-only
- [ ] The job refuses to push to `main`
- [ ] The artifact upload is retained as a fallback
- [ ] The commit message names the run id and head SHA
- [ ] Verified end-to-end by dispatching it on a branch with a deliberate pixel change

### Context

Came out of #287 (the light flip), where all nine baselines needed regenerating and the loop could not be closed from the session that ran it.

skizzy1986:
Closes #288

## What changed and why

`Update Visual Baselines` now commits the regenerated screenshot baselines to the dispatched branch instead of stopping at an artifact — removing the manual download-unzip-commit step that #287 ran straight into. A *remote* agent session can't close that loop: the GitHub MCP tools expose no artifact-download call and there's no Docker daemon in the sandbox. (A local machine with Docker Desktop **can** reproduce the pinned image — verified during this PR, byte-for-byte against CI — so the constraint is on remote sessions specifically. The stronger argument stands regardless: it's the step most likely to be skipped, because Visual isn't a required check and a design PR is already green by the time you get there.)

### How it's built

Regeneration and the commit are **two jobs on purpose**. The PNGs can only be produced inside the pinned `mcr.microsoft.com/playwright:vX.Y.Z-noble` image, but pushing from inside it is fragile — `actions/checkout` falls back to a REST tarball when it can't use git, leaving no `.git` to commit into. This was a prediction when the PR was written; it is now **measured**: `git rev-parse HEAD` prints nothing usable inside that container. So the container job is unchanged, and a new `Commit Visual Baselines` job on a plain runner consumes the artifact — which stops being vestigial and becomes the handoff, while remaining the escape hatch if a push is ever refused.

Other guard rails: sanity-gates on at least one PNG before committing (the regeneration step is `continue-on-error`, so a broken upstream run must not wipe the baselines), no-ops cleanly when nothing moved, `git add` scoped to the snapshots dir, fast-forward push only — never `--force`.

This deliberately does **not** decide whether the new screenshots are *right*. That judgement stays on the PR, reading the committed PNG diff.

## ✅ Verified end-to-end

Run on a throwaway branch with a deliberate geometry change (extra copy on the login screen). The full loop:

1. **Reject** — `Playwright Visual` *fails* against the old baselines ([run 369](https://github.com/skizzy1986/erg-dashboard/actions/runs/32808546989))
2. **Regenerate + commit** — dispatch produces commit `e818b63` ([run 370](https://github.com/skizzy1986/erg-dashboard/actions/runs/32808552925))
3. **Accept** — the push retriggers CI and `Playwright Visual` *passes* against the new baselines ([run 371](https://github.com/skizzy1986/erg-dashboard/actions/runs/32808657559))

| DoD | Evidence |
|---|---|
| Commits and pushes to the dispatched branch | `e818b63` |
| Authored by the bot | `github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>` |
| Commit message names run id + head SHA | `test: regenerate visual baselines (run 32808552925)`, body cites `2f8432808d7d…` |
| Touches only snapshot PNGs | **exactly one file** — `login-visual-linux.png`, 10631 → 11823 bytes |
| Push retriggers the workflow | run 371 spawned 4s after the push; a `GITHUB_TOKEN` push spawns nothing |
| `contents: write` scoped to one job | container job's log shows `GITHUB_TOKEN Permissions: Contents: read` |
| Refuses `main` | both baseline jobs report *skipped* on every `pull_request` event |
| Artifact upload retained | attached to every dispatch run |

**No feedback loop:** on the retriggered run both baseline jobs report *skipped*, so the bot's own push cannot re-enter the commit job.

## Prerequisite (already done)

The job pushes with `VISUAL_BASELINE_PAT` — a fine-grained PAT, this repo only, Contents: read+write — so the push retriggers `e2e-web.yml`. A `GITHUB_TOKEN` push never retriggers a workflow (see `dependabot-maintenance.yml`). The workflow falls back to `GITHUB_TOKEN`, so it stays correct without the secret; it just wouldn't re-run CI.

One caveat now documented in the workflow header: the retrigger arrives via `pull_request`, not `push` (`on.push` is main-only), so the branch needs an **open PR** for the re-run. That's the normal case — baselines get regenerated on a branch already under review.

## Checklist

- [x] CI is green (lint, test, build) — plus 832 tests locally
- [x] Tests cover the change, or I've noted why they don't — **workflow YAML has no unit-test surface**; validated by YAML parse, `bash -n` on the push script, and the end-to-end run above
- [x] No unexpected console errors in the browser — no application code touched
- [x] Visual change: none — baselines untouched by this PR
- [x] Stays inside the property boundary — CI config and comments only

### DIFF
--- a/.github/workflows/e2e-web.yml
+++ b/.github/workflows/e2e-web.yml
@@ -19,9 +19,29 @@ name: E2E — Web
 # For the visual suite that pinning is load-bearing in a second way: glyph
 # rasterisation differs between images, so a baseline generated anywhere other
 # than this exact image fails forever. That is what the `Update Visual
-# Baselines` job is for — run it from the Actions tab when a design change is
-# meant to move pixels, download the artifact, commit the PNGs. A bump of
-# @playwright/test can also require it.
+# Baselines` job is for — dispatch it from the Actions tab on the branch whose
+# design change is meant to move pixels, and it regenerates the PNGs and commits
+# them to that branch (#288). A bump of @playwright/test can also require it.
+#
+# Regenerating and committing are deliberately two jobs. The PNGs can only be
+# produced inside the container, but pushing from inside it is fragile —
+# actions/checkout silently falls back to a REST tarball when it cannot use git,
+# leaving no .git to commit into — so `Commit Visual Baselines` runs on a plain
+# runner and consumes the artifact. The artifact upload is therefore load-bearing
+# now, and is still the escape hatch if the push is refused.
+#
+# `contents: write` is scoped to that one job; everything else here is read-only,
+# and it refuses to push to main. It pushes with VISUAL_BASELINE_PAT (a
+# fine-grained PAT, this repo only, Contents: read+write) so that the push
+# retriggers this workflow and Playwright Visual re-runs against the new
+# baselines. Without that secret it falls back to GITHUB_TOKEN, which still
+# commits but never retriggers a workflow (see the note in
+# dependabot-maintenance.yml): the PNGs land, but the Visual check keeps showing
+# the previous run until something else pushes.
+#
+# What this does NOT do is decide whether the new screenshots are RIGHT. That
+# judgement belongs on the PR, reading the committed PNG diff — the job only
+# removes the clerical step of moving files.
 #
 # Path filtering is job-level (`changes` job + `if:`), not workflow-level
 # `on.paths` — see the note in ci-web.yml for why (required-check safety).
@@ -33,9 +53,14 @@ on:
   workflow_dispatch:
     inputs:
       update_baselines:
-        description: 'Regenerate the visual baselines and upload them as an artifact'
+        description: 'Regenerate the visual baselines and commit them to this branch'
         type: boolean
         default: false
+
+# Read-only floor for the whole workflow, so the one `contents: write` below is
+# an explicit, auditable exception rather than an inherited repo default.
+permissions:
+  contents: read
 
 concurrency:
   group: e2e-web-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


skizzy1986:
Internal change — no release note. This PR only touches CI workflow config (`.github/workflows/e2e-web.yml`) and a test-infra comment (`web/e2e/visual.spec.js`); no application code or user-visible behavior changed.

---

Issue #291, opened 2026-08-25
Visual baselines do not move when THEME.bg changes, despite the built bundle being correct


skizzy1986:
### What needs doing

Changing `THEME.bg` in `web/src/constants/theme.js` from `#08080d` to a visibly different colour produces **zero pixel change** in all nine visual baselines when regenerated in the pinned Playwright container — even though the built bundle provably contains the new value and provably reads it.

`error-fallback-visual-linux.png` is the clearest case. `ErrorFallback.jsx:16` paints `background: THEME.bg` across `minHeight: 100vh`, and that baseline is ~99% `#08080d`. Change the token and the PNG should be ~99% the new colour. It isn't — it comes back byte-identical.

This matters beyond the cosmetic: it means at least one colour token is not reaching the surface it appears to paint, and the visual suite is silently insensitive to that class of change. Both are directly relevant to **#183** (the palette work) — moving values that nothing reads would produce a green suite and no visual change.

Found while verifying #288 (PR #289). It is unrelated to that change; the baseline-commit job works correctly and was verified with a geometry stimulus instead.

### Evidence

Reproduced on **three independent runs** — GitHub Actions twice, and a local Docker Desktop run of the same `mcr.microsoft.com/playwright:v1.62.1-noble` image (linux/amd64) that had already been proven to reproduce CI byte-for-byte on unmodified `main`.

| Check | Result |
|---|---|
| Source at build time | `bg: '#3d0812'` confirmed present |
| `dist/` contents | `3d0812` ×1, `08080d` ×49 |
| Served to the browser over HTTP | `index-*.js`, 1379809 bytes, `3d0812` ×1 |
| `--update-snapshots` in the container | 9 passed, **PNGs unchanged**, twice |
| Regenerated PNG, decoded | 1280×800, **99.3% `#08080d`** — the old colour |
| Baseline PNG checksum | identical before and after |

So the browser genuinely rendered and screenshotted that screen, and produced the old colour.

### The build is not at fault — this is the useful narrowing

Built with a unique sentinel (`bg: '#abcdef'`) and inspected the output. The theme object survives minification as a live object:

```js
var V={bg:`#abcdef`,surface:`#1a1a2e`,raised:`#2a2a48`,fie…
```

and ErrorFallback reads it as a property, not an inlined constant:

```js
…justifyContent:`center`,gap:14,background:V.bg,color:V.text,fontFamily:`ui-sans-serif, …
[...]

Issue #279, opened 2026-08-24
fix(design): `constants/ui.js` duplicates seven THEME hexes as raw literals — they won't follow the light flip


skizzy1986:
Found during #248 (PR #277) and deliberately left out of scope there — touching it moves hex values, which would have broken that PR's core "no value changed" criterion.

`web/src/constants/ui.js` maps nine session types to colours as **raw hex literals**, seven of which duplicate values that also live in `THEME`:

`#00d4ff` · `#ffd700` · `#ff6b35` · `#ff2d55` · `#a78bfa` · `#34d399` · `#f472b6` · `#2dd4bf` · `#3a3a4a`

## Why this is a real problem, not a style nit

`CLAUDE.md` and `.claude/skills/erg-context.md` both state the rule plainly: *read every colour from `THEME` — never type a raw hex; a literal is a defect even when it currently matches.*

The consequence is concrete. `LogEntry` and `WorkoutItem` colour their session-type accents from this `C` map, so **when the light flip lands, those colours will not move.** Every other colour in the app will change and these nine will stay dark-theme — a visible, hard-to-trace regression.

Related: `web/src/components/LogEntry.jsx:331` has an orphan `'#ffaa44'` with no `THEME` token at all, so it will survive the flip as a dark-theme artefact too.

## Scope

- Point the `C` map at `THEME` keys instead of literals.
- Decide what to do with `#ffaa44` — give it a token or replace it with an existing role.
- Values may change here (that is the point), so this needs a screenshot/baseline review, unlike #248.

This is #183's lane.

skizzy1986:
**Triage**

**Priority: High** — genuine regression risk, not a style nit: `constants/ui.js` and the orphan hex in `LogEntry.jsx:331` will silently stay dark-themed once the light flip ships, while every other colour in the app moves. That's a visible, hard-to-trace bug for whoever hits it post-flip.

**Duplicates:** none — distinct from #248 (closed, literal-token rename only) and #278 (alias-map identifiers, no hex values). Note: already referenced as closed-by PR #287 (the light-flip PR), so this may already be in flight — worth confirming status there before starting fresh work.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #294, opened 2026-08-25
chore(sessions): give every Claude Code session its own worktree

skizzy1986:
## Why

Sessions launched in the same directory share one working tree and one `HEAD`, with no isolation between them.

On **2026-08-25** three concurrent sessions — one on #279, one on #287, one on #291 — collided in this repo. In forty minutes the tree changed branches four times (`fix/residual-theme-hexes` → `main` → `claude/mobile-app-design-adoption-92zgfw` → `fix/291-visual-comparator-threshold`), and one session's uncommitted work surfaced in another session's `git status` while a subagent was mid-read. Two of those sessions were writing to the same files. Nothing was lost, but that was luck rather than design.

`EnterWorktree` activates on **project instructions**, so writing the convention into `CLAUDE.md` is what makes future sessions self-isolate instead of discovering the collision the hard way.

## What changed

| File | Change |
|---|---|
| `CLAUDE.md` | New "Concurrent sessions: one worktree each" subsection under Change Procedure, plus a Safety Constraints bullet |
| `.claude/settings.json` | `worktree.baseRef: "fresh"` — a new worktree branches from `origin/main`, matching step 1 of the Change Procedure |
| `.gitignore` | Ignore `.claude/worktrees/` so per-session checkouts aren't untracked noise |

The doc covers the four traps that actually bite, each hit while writing this PR:

- **`EnterWorktree` always cuts a _new_ branch.** Continuing an open PR needs `git worktree add <path> <branch>` then `EnterWorktree` with `path`.
- **A switch doesn't carry uncommitted changes across** — they stay in the tree you left. Commit or stash first.
- **One branch, one worktree.** Git refuses to check the same branch out twice, and that refusal is the constraint doing the real work.
- **`npm install` in `web/` comes first.** A worktree is a fresh checkout with no `node_modules`, and `.husky/pre-push` runs the full vitest suite — so without it you cannot push at all.

## Verification

- `npx vitest run` — **832 passed, 67 files** (run in the worktree, via the pre-push hook; not bypassed)
- `settings.json` parses and all four hook groups are intact
- `.claude/worktrees/` confirmed matched by the new rule (`.gitignore:30`)
- Working tree clean after install + test run, so `node_modules/` and `test-results/` are already ignored

Docs and config only — no source touched, so `lint` / `format:check` / `test` scope (`web/src`, `web/scripts`) is unaffected.

## Note for review

`.claude/settings.json` has two hooks pinned to `/home/user/erg-dashboard`, a Linux path that doesn't exist on this machine — the `PostToolUse` eslint hook and the `Stop` uncommitted-changes warning. Both are silently dead here. The `Stop` one is the warning that would have flagged the uncommitted work left sitting in the shared tree. Out of scope for this PR; worth its own issue.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,4 +1,5 @@
 ﻿{
+  "worktree": { "baseRef": "fresh" },
   "hooks": {
     "SessionStart": [
       {
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,7 @@ web/.design-sync/node_modules
 # written by whatever Deno happens to be on a dev machine is noise, not a
 # reproducibility guarantee.
 deno.lock
+
+# Per-session git worktrees (EnterWorktree). Concurrent Claude Code sessions each
+# get their own checkout here, so two sessions cannot race on one working tree.
+.claude/worktrees/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -515,10 +515,41 @@ Rebase early and often; a branch that tracks `main` closely rarely conflicts. Ne
 to `#build` in Slack. Check `#build` to confirm a change landed correctly
 without needing to prompt Coach.
 
+### Concurrent sessions: one worktree each
+
+**Every Claude Code session works in its own git worktree.** Call `EnterWorktree`
+before the first edit — not as a rescue once a conflict appears.
+
+Sessions launched in the same directory share one working tree and one `HEAD`.
+There is no isolation between them: a `git checkout` in one session yanks the
+files out from under another mid-edit, and two sessions editing one file race on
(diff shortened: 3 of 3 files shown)


skizzy1986:
The dead-hook note under **Note for review** is now tracked as #295 — including a second bug found while writing it up: even with the path corrected, the `Stop` hook's `; if [ "$CHANGES" -gt 0 ]` runs against an unset variable when the `cd` fails, so it needs hardening too, not just a path swap.

skizzy1986:
Internal change — no release note. This PR only touches developer/process files (`CLAUDE.md`, `.claude/settings.json`, `.gitignore`) to give each Claude Code session its own git worktree; no app code or user-facing behavior changed.

---

PR #296, opened 2026-08-25
fix(hooks): revive the two hooks that were failing silently with exit 0

### LINKED ISSUE #295: fix(hooks): two hooks are pinned to a Linux sandbox path and fail silently with exit 0
Two of the four hook groups in `.claude/settings.json` `cd` to `/home/user/erg-dashboard` — a path that does not exist on Scott's Windows machine. Both have been dead for the entire life of the repo on this hardware, and neither reports a failure.

## The two dead hooks

| Group | Intent | Status |
|---|---|---|
| `PostToolUse` (matcher `Edit\|Write`) | run `eslint --max-warnings=0` on changed `.js`/`.jsx` after every edit | dead |
| `Stop` | warn when files are left changed and uncommitted at end of turn | dead |

The other two are fine: `SessionStart` uses `$CLAUDE_PROJECT_DIR`, and `PreToolUse` uses a repo-relative path.

## Why it is silent — the important part

Reproduced directly:

```
$ bash -c 'cd /home/user/erg-dashboard && CHANGES=$(git status --short 2>/dev/null | wc -l | tr -d " "); if [ "$CHANGES" -gt 0 ]; then echo "[HOOK] ..."; fi'
bash: line 1: cd: /home/user/erg-dashboard: No such file or directory
bash: line 1: [: : integer expected
exit=0
```

**Exit code 0.** The hook reports success, so nothing surfaces the failure — it looks identical to "ran fine, nothing to report."

There is a second, independent bug in the `Stop` hook visible in that trace. The command is `cd X && CHANGES=$(...)` followed by `; if [ "$CHANGES" -gt 0 ]`. The `&&` correctly short-circuits so `CHANGES` is never assigned, but the `;` runs the `if` anyway against an empty string — hence `[: : integer expected`. Fixing the path alone leaves this fragile; the guard should handle an unset `CHANGES`.

## Why it matters now

The `Stop` hook is *specifically* the "you have uncommitted changes" warning. On 2026-08-25 three concurrent sessions collided in this repo and left uncommitted work sitting in a shared working tree, which is what #294 addresses. That is exactly the condition this hook exists to announce, and it has never once fired here.

The `PostToolUse` hook means no edit has ever been linted at write time; every lint failure has had to wait for the pre-commit hook or CI.

## The fix already exists in this file

PR #10 ("Fix session-start hook failing on paths with spaces (Windows)") fixed this exact class of bug for the `SessionStart` hook:

```diff
- "command": "$CLAUDE_PROJECT_DIR/.claude/hooks/session-start.sh"
+ "command": "bash \"$CLAUDE_PROJECT_DIR/.claude/hooks/session-start.sh\""
```

The same treatment was never applied to the other two. They date from `0801751` ("Establish software factory..."), which was written against a Linux sandbox.

`$CLAUDE_PROJECT_DIR` is also the *right* variable now that #294 makes one-worktree-per-session standard: it follows the session into its worktree, so the hook lints the checkout being edited. A hardcoded repo path would silently operate on a different worktree — a worse failure than the current one, because it would appear to work.

## Suggested scope

- Replace both `cd /home/user/erg-dashboard` with `cd "$CLAUDE_PROJECT_DIR"`, quoted (the path contains spaces on this machine — `...\OneDrive\Desktop\Fitness Coach\...`)
- Make each hook fail loudly rather than exit 0 when its `cd` fails
- Harden the `Stop` hook's integer comparison against an empty `CHANGES`
- Consider whether `PostToolUse` should lint **the file just edited** rather than `git diff --name-only HEAD | head -5` — the current form lints an arbitrary five changed files, which may not include the one that triggered it
- Verify on Windows that both fire, ideally by checking the hooks with a deliberate lint error and a deliberate uncommitted file

Found while setting up per-session worktrees (#294), where it was flagged for a separate issue rather than folded into that PR.

skizzy1986:
Closes #295

## The bug

`PostToolUse` (edit-time eslint) and `Stop` (uncommitted-changes warning) both `cd /home/user/erg-dashboard`, a path that exists only in the original Linux sandbox. Off it the `cd` fails — and the hook still exits 0, so both have been dead on this machine for the life of the repo while reporting success.

Reproduced:

```
$ bash -c 'cd /home/user/erg-dashboard && CHANGES=$(git status --short | wc -l | tr -d " "); if [ "$CHANGES" -gt 0 ]; then echo "[HOOK] ..."; fi'
bash: line 1: cd: /home/user/erg-dashboard: No such file or directory
bash: line 1: [: : integer expected
exit=0
```

Two distinct faults in that trace. The path, and — separately — `cd X && CHANGES=$(...)` followed by `; if [ "$CHANGES" -gt 0 ]`: the `&&` short-circuits so `CHANGES` is never assigned, but the `;` runs the comparison against an empty operand regardless.

The `Stop` hook is *specifically* the warning that would have flagged the uncommitted work left in a shared checkout during the 2026-08-25 three-session collision (#294). It has never fired here.

## The fix

Both move into scripts under `.claude/hooks/`, matching `session-start.sh` and `block-secret-commit.sh`, invoked with the quoted `$CLAUDE_PROJECT_DIR` form **PR #10 already established** for `session-start.sh` — the same fix, finally applied to the other two.

`$CLAUDE_PROJECT_DIR` is now the *correct* variable, not merely the portable one: it follows a session into its worktree, which matters since #294 makes one-worktree-per-session standard. A hardcoded repo path would lint a different checkout than the one being edited — a worse failure than today's, because it would appear to work.

Beyond the path:

- **Misconfiguration exits non-zero and says so.** Lint findings stay advisory (exit 0) — a warning shouldn't block an edit, but a broken hook must never again look like a clean one.
- **`lint-edited-file.sh` lints the file named in the payload**, not the first five entries of `git diff --name-only HEAD`, which needn't include the file that triggered the hook. (This was the open design question in #295.)
- **It calls `web/node_modules/.bin/eslint` directly.** Bare `npx eslint` from the repo root can't see the project's eslint — root `package.json` declares only husky — so it downloads one on demand: **10.9.1 fetched against the 10.8.1 the lockfile pins**, once per edit.
- **The payload is parsed with `node`, not `sed`.** A Windows path arrives as `"C:\Users\..."` and only a real JSON parser unescapes it. A payload that won't parse is now *reported* rather than swallowed — I caught that silent-no-op in my own first draft, which is the same defect class this PR exists to remove.
- **`warn-uncommitted.sh` defaults its count to 0** so the comparison can't see an empty operand, and counts with `grep -c .` rather than `wc -l` (whose padding the old `tr -d ' '` was working around).

## Verification

`test-hooks.sh` — 9 cases, both happy paths and every failure mode above:

```
$ bash .claude/hooks/test-hooks.sh
ok    warn-uncommitted: missing dir fails loudly
ok    warn-uncommitted: unset var fails loudly
ok    warn-uncommitted: runs against a real repo
ok    lint: missing dir fails loudly
ok    lint: unparseable payload fails loudly
ok    lint: no file_path is a quiet no-op
ok    lint: non-JS file is a quiet no-op
ok    lint: clean file in web/src reports nothing
ok    lint: catches a real lint error

9 passed, 0 failed
```

The last case writes a file with an unused variable, runs the hook, and asserts the warning comes back — end-to-end proof the hook now lints, which the old one never did. Also `npx vitest run` via pre-push: **832 passed**.

These hooks broke silently and stayed broken for months. The test is the half that keeps them fixed.

## A correction worth recording

While writing this I claimed a third bug — that running eslint from the repo root would find no config, since the only `eslint.config.js` is `web/eslint.config.js`. **That was wrong.** ESLint 9+ resolves flat config upward from the file being linted, so the root cwd was fine. Verified directly before it reached this PR; the version-drift point above is the real defect, and it's a smaller one.

## Note

No `web/**` files touched, so `Lint & Format` / `Test & Coverage` / `Build` will path-filter to skipping. Stacks cleanly with #294 — that PR adds a `worktree` key at the top of `settings.json`, this one rewrites two hook commands further down.

### DIFF
--- a/.claude/hooks/lint-edited-file.sh
+++ b/.claude/hooks/lint-edited-file.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+# PostToolUse(Edit|Write): lint the file Claude just edited.
+#
+# Three things this gets right that the previous inline one-liner did not:
+#
+# 1. `$CLAUDE_PROJECT_DIR`, not a hardcoded /home/user/erg-dashboard. The old
+#    path does not exist outside the original Linux sandbox, so the hook `cd`
+#    failed and — because the failure exited 0 — reported success while doing
+#    nothing. It also follows the session into its worktree, which matters now
+#    that each session gets its own checkout.
+# 2. It lints the file named in the tool call. The old form linted whatever the
+#    first five entries of `git diff --name-only HEAD` happened to be, which
+#    need not include the file that triggered the hook.
+# 3. It invokes web/node_modules/.bin/eslint directly. The old form ran bare
+#    `npx eslint` from the repo root, whose package.json declares only husky —
+#    so npx cannot see the project's eslint and downloads one on demand
+#    (observed: 10.9.1 fetched against the 10.8.1 the lockfile pins). That is a
+#    network round-trip per edit and a version that drifts from the lockfile.
+#    Flat-config lookup is not the problem: eslint resolves eslint.config.js
+#    upward from the file being linted, so web/eslint.config.js is found from
+#    anywhere in the repo.
+#
+# Advisory by design: lint findings print but exit 0, so a warning never blocks
+# an edit. A misconfigured hook, by contrast, exits non-zero and says so — the
+# whole point of this fix is that a broken hook must not look like a clean one.
+set -uo pipefail
+
+root="${CLAUDE_PROJECT_DIR:-}"
+if [ -z "$root" ] || [ ! -d "$root" ]; then
+  echo "[HOOK] lint-edited-file: CLAUDE_PROJECT_DIR unset or not a directory (${root:-<unset>})" >&2
+  exit 1
+fi
+
+# The hook payload arrives as JSON on stdin. Parsed with node rather than the
+# sed approach used by block-secret-commit.sh: a Windows path arrives as
+# "C:\\Users\\scott\\..." and only a real JSON parser unescapes it correctly.
+#
+# node exits 3 if the payload will not parse, so an unreadable payload is
+# reported rather than swallowed — a silent no-op here would be the same defect
+# this hook is being fixed for.
+if ! file="$(node -e 'let s="";process.stdin.on("data",d=>s+=d).on("end",()=>{try{process.stdout.write(String(JSON.parse(s)?.tool_input?.file_path??""))}catch{process.exit(3)}})')"; then
+  echo "[HOOK] lint-edited-file: could not parse the hook payload as JSON" >&2
+  exit 1
+fi
+
+# Parsed, but no file path — e.g. a tool call this matcher does not care about.
+[ -n "$file" ] || exit 0
+
+case "$file" in
+  *.js | *.jsx) ;;
+  *) exit 0 ;;
+esac
+
+web="$root/web"
+eslint="$web/node_modules/.bin/eslint"
+
+if [ ! -x "$eslint" ]; then
+  echo "[HOOK] eslint not installed — run 'npm install' in web/ to enable edit-time linting." >&2
+  exit 0
+fi
+
+# Mirror `npm run lint`, which covers only src/ and scripts/ under web/.
+case "$file" in
+  "$web"/src/* | "$web"/scripts/*) ;;
+  *) exit 0 ;;
+esac
+
+cd "$web" || exit 1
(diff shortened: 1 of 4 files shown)


skizzy1986:
Internal change — no release note. This PR only fixes `.claude/hooks/*` dev-tooling scripts (edit-time lint hook, uncommitted-changes warning hook) and `.claude/settings.json`; no `web/**` files are touched and there's no user-visible change to the dashboard app.

---

Issue #254, opened 2026-08-23
feat: self-host Archivo and swap the DM Mono call sites (S-1a — family only)


skizzy1986:
Where the redesign visibly starts, and the cheapest real win available.

**44 call sites request `'DM Mono'` and the face is never served.** `index.html` loads no webfont and there is no `@font-face` anywhere, so every number in SplitIQ — splits, watts, heart rate, e1RM, the entire live erg readout — renders in the Courier fallback. The monospace-numeral identity the code asks for has never actually been seen. A meaningful part of how the app currently looks is an accident, not a decision.

## Scope — family only

Split the brief's S-1 in two. This is **S-1a**:

- Self-host **Archivo, 500+ weights only** (`conventions.md` rules out lighter on light grounds)
- Add `FONT` tokens
- Swap the 44 `'DM Mono'` references
- Keep a real numeral face for figures — `DESIGN_BRIEF.md` §6.1's argument for splitting a UI sans from a numeral face still stands even though the families it named were superseded
- **No size changes**

**S-1b is separate and later**: applying the TYPE scale reassigns ~419 declarations and moves 229 `fontSize: 9` sites up to 12/14. That is a layout-shifting change of the same order as the spacing pass and must not ride along with a font swap.

## Two coupled follow-ups

- **`cfg.extraFonts` is not in `config.json` today.** Fixing the app does not fix generated designs — the design bundle ships its own font closure. Without it, every generated design keeps rendering in the fallback after the app is correct, and nothing warns you.
- **Capacitor serves from `file://`** — self-hosted only, no Google Fonts request.

## Acceptance

- [ ] `grep -ro "DM Mono" web/src/ | wc -l` goes 44 → 0
- [ ] A large, uniformly type-only screenshot diff, reviewed in the PR
- [ ] `npm run size` delta stated in the PR body (font payload)
- [ ] Lighthouse performance does not drop below its warn floor

Property boundary: touches `fontFamily` only — no colour hexes, no box metrics. Collides with neither #183 nor the spacing pass.

Blocked by the baseline harness — a font swap should produce a large, expected, entirely reviewable diff, which is the perfect first proof that the harness works.

skizzy1986:
**Triage**

[...]

PR #298, opened 2026-08-25
docs(design): adopt the 2026-08-25 design drop — desktop deep dives, Archivo revert, five-chip rail

skizzy1986:
Commits the 2026-08-25 design drop into the design-owned half of `web/.design-sync/` (per `ownership.json`), verbatim — all 15 files verified byte-identical to the drop after the pre-commit hook ran.

## What the drop changes

**Two new desktop artboards** — `desktop-progress.html`, `desktop-body.html`. Eight artboards now, not six.

**The platform split is now a decision, not a gap.** Mobile does the doing; desktop does the understanding. Every live surface — prescription, watt band, set logger, rest timer, sRPE — is mobile-only. **Train has no desktop counterpart**, which is why the desktop nav rail carries five chips rather than a dimmed sixth. This directly rescopes Phase C: `PrescriptionCard`, `SessionLibrarySheet`, `BandGauge`, `SetRow`, `LiftCard`, `RestTimer`, `SessionComplete` and `SessionBar` are each built **once, for the phone**.

**Typeface reverted to Archivo + IBM Plex Mono.** `STATE_OF_PLAY.md` §4.1 had re-decided this as Plex Sans on metric-compatibility grounds; design overruled it. §4.1 now carries the supersession inline so a code session cannot act on the overruled decision. #254 stands as written.

**`conventions.md`** gains the full eight-size type ladder (9 is the floor, nothing renders below it), the six-band zone table, the desktop icon decision and its reasoning, and the rule that *a state's encoding must not be a colour another role already owns* — with the Resting HR pane as the worked example of that collapsing.

**`splitiq-load.js`** gains `zoneTable()`, `paceZones()`, `ergOf()`/`splitOf()`, `strengthSeries()`/`e1rm()` and `readinessSeries()`. Pure additions. This is the module `HANDOFF.md` §5 asks the app to mirror as one `lib/load.js`.

**`HANDOFF.md`** §2's inventory absorbs the desktop rows into the single table; the separate desktop table is gone. Readiness NO DATA is now drawn (desktop Body) — the mobile card still needs it.

## Repo-side consequences

- **`DESIGN.md`** is the front door and was stale on three counts: five artboards are now eight, the platform split is a hard rule that decides what gets built where, and `splitiq-load.js` / `splitiq-light-tokens.css` were never listed at all.
- **`check:zones` would have failed this PR.** Its CP detection required a literal `~`. The mirror writes `CP ≈ 205W` and is committed verbatim, so the guard read a document whose bands are correct as quoting no CP and failed it — and the mirror cannot be hand-edited to suit the guard. The class now takes both approximation signs. The guard goes from 3 published bands checked to 4, so the new mirror's band is now covered rather than skipped.

## Verification

- `npm run check:design-sync` — passes (47 files, all owned, nothing design-owned pushed back)
- `npm run check:zones` — passes, 4 bands across 284 tracked markdown files
- `npm run lint` — clean
- `npm test` — 832 passed / 67 files
- All 15 design-owned files verified byte-identical to the drop post-commit

## Note on diff size

Three mobile artboards (`today`, `train`, `progress`) differ **only by asset UUID** — the exporter churns them even when the design is unchanged. `body`, `coach` and `desktop-overview` carry real content changes.


skizzy1986:
Internal change — no release note. This PR only touches the design-sync documentation set (`web/.design-sync/`) and the `check-zone-bands.mjs` guard script; no `web/src` app code changed, so there's nothing user-visible to ship.

---

PR #302, opened 2026-08-25
fix(coach): let the Sentry tracing headers through the function CORS preflight

### LINKED ISSUE #301: fix(coach): Sentry tracing headers fail the coach-chat CORS preflight — Coach tab is dead
## Symptom

Sending a message in the Coach tab shows `Failed to fetch (swdrueaserjzhuxnzmeu.supabase.co)` in the error banner. Sentry: `ERG-DASHBOARD-3`, thrown at `web/src/hooks/useCoach.js:212`.

## Diagnosis

The edge logs for the failing send (2026-08-25T07:03) show the preflight and nothing else:

```
07:03:50.586  POST    /rest/v1/coach_messages        201   <- user message insert, fine
07:03:52.080  booted (68ms)
07:03:52.109  OPTIONS /functions/v1/coach-chat       200   <- preflight
07:05:07.096  shutdown                                     <- idle timeout, POST never arrived
```

The POST is never sent. That is the signature of a rejected CORS preflight.

`browserTracingIntegration()` (added in #271) attaches `sentry-trace` and `baggage` to every request matching `tracePropagationTargets` in `web/src/utils/sentry.js` — a list that deliberately includes the Supabase origin. Neither header is CORS-safelisted, so both land in `Access-Control-Request-Headers`. `coach-chat` answers with a fixed allow-list that does not contain them:

```
$ curl -X OPTIONS .../functions/v1/coach-chat \
    -H 'Access-Control-Request-Headers: authorization,baggage,content-type,sentry-trace'
access-control-allow-headers: authorization, x-client-info, apikey, content-type
```

Preflight fails, the browser never sends the POST, `fetch` rejects with `TypeError: Failed to fetch`.

Supabase's own gateway (`/rest`, `/auth`) reflects request headers, which is why every other call in the app kept working and only the hand-rolled function list broke.

`vitals-sync` has the identical allow-list and is also browser-called (`supabase.functions.invoke` in `useVitalsSync.js`), so it carries the same bug.

## Fix

Add `sentry-trace, baggage` to `Access-Control-Allow-Headers` in both functions, plus `Access-Control-Allow-Methods` and a `Access-Control-Max-Age` (the cold-start OPTIONS measured 1112 ms, paid on every send without it).

Note the deployed `coach-chat` is **version 12**, which predates the Sentry instrumentation in the repo — the fix needs an actual function deploy, not just a merge.

skizzy1986:
Fixes #301

## What was wrong

The Coach tab returned `TypeError: Failed to fetch` on every send. The edge logs for the failing attempt show the shape of it:

```
07:03:50.586  POST    /rest/v1/coach_messages   201   <- the user message insert, fine
07:03:52.080  booted (68ms)
07:03:52.109  OPTIONS /functions/v1/coach-chat  200   <- the preflight
07:05:07.096  shutdown                                <- idle timeout; the POST never arrived
```

No POST, anywhere. The browser rejected the preflight and therefore never sent the real request.

`browserTracingIntegration()` (#271) attaches `sentry-trace` and `baggage` to every request whose URL matches `tracePropagationTargets`, and that list deliberately includes the Supabase origin. Neither header is CORS-safelisted, so both land in `Access-Control-Request-Headers` — and `coach-chat` answered with a fixed allow-list naming neither:

```
$ curl -X OPTIONS .../functions/v1/coach-chat \
    -H 'Access-Control-Request-Headers: authorization,baggage,content-type,sentry-trace'
access-control-allow-headers: authorization, x-client-info, apikey, content-type
```

Supabase's own gateway reflects request headers back, which is why `/rest` and `/auth` kept working throughout and only the hand-maintained list broke.

## What changed

- `coach-chat` and `vitals-sync` now allow `sentry-trace` and `baggage`. `vitals-sync` had the identical list and is browser-called through `supabase.functions.invoke` in `useVitalsSync.js`, so it was one Sync click from the same failure.
- Both set `Access-Control-Allow-Methods` and `Access-Control-Max-Age`. Without a max-age the browser re-preflights every message, and that cold-start OPTIONS measured 1112 ms.
- `CLAUDE.md` records the trap next to the two existing silent-Sentry failure modes: the tracing headers are a constraint on every hand-rolled edge-function CORS list, not just a Sentry detail.

## Deploy note

**Merging this does not fix production.** The deployed `coach-chat` is version 12, which predates even the Sentry instrumentation already in the repo — the functions are deployed by hand, not by CI. Both functions need an actual deploy after merge.

## Testing

- `npm test` passes (pre-push hook).
- No `web/src` files touched, so lint/format/coverage scope is unchanged — CI does not lint the Deno functions at all.
- Preflight behaviour verified against the live endpoint with `curl` before and after.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -284,6 +284,15 @@ when Sentry looks dark:
 `browserTracingIntegration()` is **not** a default integration in the browser SDK;
 `tracesSampleRate` does nothing without it. There is a test asserting this.
 
+It also has a **blast radius outside Sentry**: `tracePropagationTargets` in
+`web/src/utils/sentry.js` includes the Supabase origin, so the SDK attaches
+`sentry-trace` and `baggage` to every call to it. Neither header is CORS-safelisted,
+so both appear in the preflight. Supabase’s own gateway (`/rest`, `/auth`) reflects
+request headers and is unaffected, but **every hand-rolled edge-function
+`Access-Control-Allow-Headers` must list them** or the browser fails the preflight and
+never sends the request — a bare `TypeError: Failed to fetch` on the client with only
+the OPTIONS in the edge logs. This silently killed the Coach tab (#301).
+
 Session Replay is deliberately **not** enabled — it costs ~35-50 KB gzip against
 ~40 KB of headroom under the 400 KB budget in `web/scripts/check-bundle-size.mjs`.
 
--- a/supabase/functions/coach-chat/index.ts
+++ b/supabase/functions/coach-chat/index.ts
@@ -10,9 +10,25 @@ import { captureFunctionError } from '../_shared/sentry.ts';
 
 const FN = 'coach-chat';
 
+// `sentry-trace` and `baggage` are not decoration: browserTracingIntegration
+// attaches them to every request whose URL matches tracePropagationTargets in
+// web/src/utils/sentry.js, and that list deliberately includes this Supabase
+// origin. Neither is CORS-safelisted, so both land in the preflight’s
+// Access-Control-Request-Headers — and a preflight that does not allow them
+// fails in the browser, which then never sends the real request at all. The
+// symptom is a bare `TypeError: Failed to fetch` on the client with nothing but
+// the OPTIONS in the edge logs. Supabase’s own gateway (/rest, /auth) reflects
+// request headers back, which is why every other call kept working and only
+// this hand-maintained list broke. Anything the frontend attaches to outgoing
+// requests has to be listed here.
 const CORS_HEADERS = {
   'Access-Control-Allow-Origin': '*',
-  'Access-Control-Allow-Headers': 'authorization, x-client-info, apikey, content-type',
+  'Access-Control-Allow-Headers':
+    'authorization, x-client-info, apikey, content-type, sentry-trace, baggage',
+  'Access-Control-Allow-Methods': 'POST, OPTIONS',
+  // Without this the browser re-preflights every message, and on a cold isolate
+  // that OPTIONS has measured at ~1.1s before the real request even starts.
+  'Access-Control-Max-Age': '86400',
 };
 
 const MAX_CONTEXT_CHARS = 8000;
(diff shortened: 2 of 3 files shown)


skizzy1986:
## Deployed and verified

`coach-chat` **v13** and `vitals-sync` **v4** are live from this branch.

Preflight now answers with what the browser actually asks for:

```
access-control-allow-headers: authorization, x-client-info, apikey, content-type, sentry-trace, baggage
access-control-allow-methods: POST, OPTIONS
access-control-max-age: 86400
```

`coach-chat` verified end-to-end with a real POST — 200, `text/event-stream`, text deltas, `message_stop`. It boots fine with `_shared/sentry.ts`, which had never actually run in production on any function before this.

`vitals-sync` is verified at the preflight and gateway only; I did not invoke it, since that writes vitals and calls Google. Pressing **Sync** in the app is the natural confirmation.

## One thing found on the way in

The repo's `vitals-sync/index.ts` was **behind** production, not ahead. The live v3 selects the real column names (`rhr_bpm, hrv_ms, sleep_hours, bodyweight_kg`) and syncs four extra Google-Health data types; the committed copy still selected `rhr, hrv, sleep, bodyweight` — which do not exist — and dropped steps, distance, active minutes and calories.

Deploying the repo version as-is would have quietly regressed the on-demand sync. `d0e67ed` transcribes the deployed body back into git so the two agree, and v4 was deployed from exactly that file.

Worth knowing that nothing would have caught this: `npm run lint` / `format:check` cover `web/src` and `web/scripts` only, so `supabase/functions/**` has no CI coverage of any kind. The same blind spot is why the CORS bug shipped.

skizzy1986:
**Release notes draft — Fixed**

Fixed the Coach chat tab silently failing on every message send ("Failed to fetch"). A CORS preflight change needed after recent error-tracking updates wasn't applied to two of the app's backend functions, so the browser blocked outgoing requests before they ever reached the server. Coach chat and background health-data sync are both back to working normally.

_Note: this PR also updated internal documentation — no separate note needed for that part._

---

Issue #278, opened 2026-08-24
refactor: unwind the local `C` colour-word alias maps before the var(--color-*) seam


skizzy1986:
Fallout from #248 (PR #277). Not a defect — behaviour is identical — but it will trip up #250.

Several views define a local `C` alias map whose **keys are still colour words** while its values now point at role-named `THEME` keys. Those keys are local variables, not `THEME` tokens, so #248's "no colour-named key remains" grep correctly ignored them and the rename left them alone.

The result reads badly. `web/src/views/ErgLiveView.jsx:13`:

```js
const C = {
  accent: THEME.positive,   // C.accent is GREEN
  cyan: THEME.accent,       // C.cyan is the accent colour
  gold: THEME.caution,
  err: THEME.critical,
};
```

`C.accent` is not the accent; `C.cyan` is. `web/src/views/CoachView.jsx:5` has the same shape (`cyan: THEME.accent`), and the mobile views carry similar maps.

## Why it matters

These maps are exactly the indirection layer the `var(--color-*)` seam (#250) has to unwind. An inverted `accent`/`cyan` pair is a landmine for whoever does that work — the obvious mechanical substitution produces wrong colours.

## Scope

- Rename the map keys to match the roles they hold, or drop the maps and read `THEME` directly where the indirection buys nothing.
- Pure identifier work: **no hex may change**, and the same first criterion applies — every hex on both a `+` and a `-` line.
- Property boundary: colour identifiers only. No sizes, no type.

Do this before #250, not during it.

skizzy1986:
**Triage**

**Priority: High** — not a live bug (behaviour is identical today), but it's explicitly a landmine for #250: the inverted `accent`/`cyan` naming in `ErgLiveView.jsx`/`CoachView.jsx` means a mechanical `var(--color-*)` substitution during #250 would silently swap colours. Sequencing matters — this needs to land before #250 starts, so it should be prioritized ahead of that work despite being cosmetic in isolation.

**Duplicates:** none — distinct from #279 (raw hex literals, not identifier naming). Note: already referenced as closed-by PR #287, so it may already be addressed there — worth checking before starting fresh work.

---
_Generated by [Claude Code](https://claude.ai/code)_

skizzy1986:
**Absorbed by #287 — no separate work needed.**

[...]

Issue #260, opened 2026-08-23
feat: draw the pending state for training load on Progress


skizzy1986:
From `web/.design-sync/ISSUES-load-states.md` (Issue 1), written by the design side and landed in `9b896df`. One of two load states no screen currently draws. Both were introduced by the switch from the static `DAILY_TSS` seed to the live `useTSSHistory` hook.

## Problem

`App.jsx` distinguishes a slow first read (`loadPending`) from genuinely absent data, and #196 established that pending must **not** flash an outage line and then replace it with real numbers. The Progress screen has no pending treatment — it assumes CTL/ATL/TSB and the session bars always arrive.

## What's needed

A pending state for the Load pane covering:

- the three-up stat row (fitness / fatigue / form)
- the session-bar chart and its fatigue ribbon
- the derived caption (which states a comparison and cannot be written until the data lands)

## Constraints

- No outage copy, no error styling, no zeroes standing in for real values.
- The layout must not reflow when the numbers arrive — reserve the space.
- Fatigue/fitness/form keep their accent colours; the placeholder is the numerals, not the labels.
- Silence beats a spinner per element — one quiet indication for the card.

## Acceptance

- [ ] With `loadPending: true` the pane renders its full layout, no numbers, no outage language
- [ ] Nothing shifts position when data arrives
- [ ] Derived captions are absent while pending rather than partially written
- [ ] Covered by a test asserting pending ≠ unavailable (mirrors `OverviewView.test.jsx:154`)

---

### Verified against the code (2026-08-23)

The spec was written on branch `fix/s2c-strength-prs-cp`; its references still hold on `main`:

- `App.jsx:185` — `const { data: tssHistory, isLoading: loadPending } = useTSSHistory();`
- `App.jsx:195-196` — `loadAvailable = latest != null`, so `loadUnavailable = latest == null && !loadPending`, exactly as the spec states it
- `OverviewView.test.jsx:153-154` — the existing pending-≠-unavailable test to mirror, with the `#196` rationale in a comment

### Sequencing

`HANDOFF.md` §5 puts the load states at step 2, immediately after the token seam (#250, #251). The Progress screen itself is one of the five redesign destinations, so this lands with that work rather than against the current UI. The reference design is `web/.design-sync/designs/progress.html` — though note the pending state is one of the **undrawn** states listed in `designs/README.md`, so this issue is the specification and the drawing does not exist yet.

skizzy1986:
**Triage**

[...]

Issue #251, opened 2026-08-23
feat: flip SplitIQ to the light palette and raise the Lighthouse a11y floor


skizzy1986:
Final palette step. `web/.design-sync/HANDOFF.md` §1. By this point only the **values module** changes — the seam and the rename have already landed.

## Corrected: dark is retained, not deleted

`HANDOFF.md` §1 says *"Dark: dropped. Light is the only theme"* and its acceptance list says *"Dark values deleted, not commented out"*. **The design project's `conventions.md`, revised 2026-08-22, overturns that in review**: dark is retained as a second theme — the erg room is dark at 5am and the live screen wants it — with light primary.

Both are design-side documents; `conventions.md` is the later one. Consequences:

- The `data-theme` attribute **stops being the no-op** §1 describes and becomes load-bearing.
- A `[data-theme='dark']` block eventually ships. Not necessarily in this PR, but the values must survive to make it possible.
- **Keep the dark values.** Do not delete them.

## The ground colour is settled: `#bcc5dd`

Not by preferring one document over another — the artboards settle it empirically:

- `#c3cade` appears **only** as the `--color-bg` declaration in each screen's helmet (10 occurrences, all the token definition itself).
- `#bcc5dd` is what the screens actually **paint** — the 390×844 device frame, the 1080px desktop frame, panel grounds.

The token is declared and never consumed. **Fix this at source before wiring the seam**: the moment `--color-bg` is genuinely consumed, every screen shifts to a colour no design was reviewed against.

## Still needed: nine token values

The handoff supplies eleven:

```
bg (see above) · surface #ffffff · border #c8cee0 · text #1c1e2a · muted #43485a
accent #0a7093 · positive #10795a · caution #8a6a10 · warning #a34c1c
critical #a32040 · accentAlt #5f45b0
```

All eleven are confirmed in use across the artboards. No light value is given for `raised`, `field`, `surfaceAlt`, `neutral`, `divider`, `textSubtle`, `textFaint`, `textDim`, or `accentAlt2`.

**Candidates from the artboards, for Design to confirm or correct** — these are the next most-used light values, and their roles are inferred from frequency, not stated anywhere:

| Hex | Uses | Likely role |
|---|---|---|
| `#646a80` | 80 | `textSubtle` |
| `#3f4457` | 44 | a darker ink — `raised`? |
| `#e4e7ef` | 28 | `surfaceAlt` or `divider` |
| `#4a4f63` | 15 | another ink step |
[...]

PR #305, opened 2026-08-25
docs(design): mirror the 2026-08-25 design drop, and declare its two new files

skizzy1986:
Design published a new handoff to Drive and asked that a code session mirror it into the repo, so the next session reads `web/.design-sync/` rather than going to Drive.

**Mirror-only.** Nothing outside `.design-sync/` changes. Two commits: the docs, then the compiled designs once design re-minted the URLs.

## The docs

Diffs ignore line endings — the mirror was checked out CRLF, the originals are LF, which alone accounted for ~950 apparent differing lines in `HANDOFF.md`. `core.autocrlf` normalises on commit, so the diff here is content only.

| File | Delta | What |
|---|---|---|
| `splitiq-load.js` | 89 | adds `SEASON`, `seasonPlan()`, `targetTxt()`, `ROSTER`, `rosterFor()`, `rosterCheck()`; **removes** Threshold/VO₂/Sharpener from the fixture and from `BAND_OF` |
| `HANDOFF.md` | 44 | new §4 Planning; `seasonPlan()`/`rosterCheck()`/`ROSTER` rows in §2; NavRail five→six chips; "Known divergence, 2026-08-25" in §5 |
| `conventions.md` | 42 | six-chip rail; new rule — *a verdict cell must be derived for every row, or absent* |
| `PROJECT-CONTEXT.md` | 9 | names all four desktop screens, not just Overview |
| `PIPELINE.md` | new | the handoff route codified: drop triggers, ordered steps each side, failure modes |
| `ownership.json` | +1 | declares `PIPELINE.md` |

`ISSUES-load-states.md` and `splitiq-light-tokens.css` were already current and so do not appear.

### Why `splitiq-load.js` is the one that mattered

Not just new features. The committed copy's fixture programmed Threshold, VO₂ and Sharpener erg sessions and mapped them to `AT` in `BAND_OF`. `CLAUDE.md`'s training model is pure base + strength — *"no programmed threshold/VO₂"*. Anyone extracting the load model from the old mirror would have built bands the programme does not use. The new copy strips them and reduces `BAND_OF` to UT1.

## The nine designs

Design re-minted the URLs, so the compiled designs land. **All nine are updated, not just the missing one** — the eight already committed came from the previous drop and every one differed, because design recompiled the set. The tell is the nav rail: each fresh desktop screen now carries a Planning chip where the committed copy had none, which is exactly the five→six change the docs describe. `desktop-planning.html` is new — the ninth design, and the only one never fetched.

Each verified at 0.7–1.7 MB and starting with `<!DOCTYPE html>`, per `FETCH-DESIGNS.md`'s own check. Nothing in `src/` or `scripts/` reads them.

## A contradiction between two design-owned docs

`PIPELINE.md` and `PROMPT-FOR-CODE.md` were written a minute apart and disagree on which files the repo mirrors:

- `PIPELINE.md`'s destination table marks `PROMPT-FOR-CODE.md`, `FETCH-DESIGNS.md` and `README.md` **"briefing only — not committed"**.
- `PROMPT-FOR-CODE.md` lists all nine as mirrored and *"all of them are `owner: design` in `ownership.json`"*.

**`PIPELINE.md` wins**, because `PROMPT-FOR-CODE.md` itself names it the doc to read "if a fetch fails, a doc contradicts another, or you are wondering who fixes what". So the first commit's `README.md` and `FETCH-DESIGNS.md` are removed here and `PIPELINE.md` added in their place.

That is also better on its own terms: a committed `FETCH-DESIGNS.md` is a block of URLs that are already dead — precisely the stale artefact `PIPELINE.md` warns about.

**For design:** worth settling which doc is authoritative, since `PROMPT-FOR-CODE.md` is the one a cold session reads first. `PIPELINE.md` also omits *itself* from its own destination table while listing itself as a drop trigger — it is mirrored here on the strength of `PROMPT-FOR-CODE.md`'s table.

## Why `ownership.json` had to change

It is repo-owned, and `web/scripts/check-design-sync-entry.mjs` fails on any file in `.design-sync/` it cannot classify. `PIPELINE.md` is `design` — mirrored down, never pushed back, and kept out of `config.json`'s push inputs, which is the two-way loop #269 closed.

## Three things in the drop that are stale — for design, not fixed here

Design-owned files are a read-only mirror, so these are raised rather than corrected in place.

1. **"What is still blocking" is wrong.** It says the token seam blocks everything with #250, #251, #278, #279 outstanding. All are closed; #277 and #287 are merged. The app is light on `main`. §6 item 1 is done and §6 item 2 already has #260/#261 open. The real next step is §6 item 3.
2. **#254 (self-host Archivo) is closed**, not outstanding as the typeface note implies.
3. **§5 disagrees with what shipped.** §5 prescribes extracting one `load.js`; #293's B1 concluded the opposite and shipped it — `calcTrainingLoad` was already the sole CTL/ATL/TSB implementation, so a new module was churn. The actual defect was *banding*: seven call sites across three schemes, now unified in `tsbBand()`. The principle held; the remedy did not.

The mobile/desktop divergence §5 flags (−8.9 / 16.4 / 25.3 vs −8.3 / 50.8 / 59.1) is untouched here, per the instruction not to close it by editing digits.

## Verification

- `check:design-sync` — passes, 49 files classified, nothing design-owned pushed back
- `check:zones` — passes, 4 published bands recompute against `derivePaceZones()`
- `lint` — clean
- `npm test` — 934 passed, 74 files
- `npm run build` — clean
- Committed `splitiq-load.js` verified byte-identical to the original after the pre-commit hook ran

Two local-only notes, neither reproducing in CI: `format:check` reports 172 untouched files (Windows CRLF artefact — `src/App.jsx` fails as CRLF, passes as LF; `.design-sync/` is prettier-ignored so the mirror stays verbatim), and `scripts/__tests__/visual-tolerance.test.mjs` timed out once on a cold `@playwright/test` dynamic import, passing in 1.7s warm.


skizzy1986:
Internal change — no release note. This mirrors a design-tool drop into `web/.design-sync/` (docs + pre-rendered design HTML); nothing in `src/` or `scripts/` changed, so there's no user-visible behavior to describe.

---

PR #315, opened 2026-08-27
feat(steward): add the PR-stewarding brief, and a staleness comment for it

skizzy1986:
Closes #314

## What changed and why

Adds `.claude/skills/steward/SKILL.md` — read automatically before an agent acts on CI or review events for a PR it owns — plus a `steward-staleness` job in `labeler.yml` that comments when a PR touches a file the skill cites.

Its whole value is the repo-specific facts that make an agent **confidently wrong**. Generic PR rules already cover etiquette; they cannot cover that CI runs UTC, or that a skipped job counts as passing.

Every scenario it addresses is a mistake made on 2026-08-26:

| Incident | Missing fact |
|---|---|
| Three wrong diagnoses of #310's missing checks | `pull_request` delivery lags ~10 min; a missing check is not a missing trigger |
| "PR #293 does not exist" | Unmerged PR commits are invisible to `git log \| grep '#N'` |
| "`splitiq-load.js:214-268` possibly fabricated" | The checkout was one commit behind |
| A DST bug shipped green | CI runs UTC only |

## Structure

Five tables, keyed to **when the fact bites** rather than by topic: before you trust the check list · when a check is red · when every check is green · when you push a fix · when you report back.

The load-bearing section is *when every check is green*, because that is where green misleads. A job skipped by an `if:` reports `skipped`, which branch protection counts as **passing** (`ci-web.yml:3-7`) — so a PR touching only `supabase/functions/**` passes `ci-web` with nothing having run. Combined with `coach-chat` and `vitals-sync` having no tests at all (#312), a fully green PR there is unverified. The skill says so in those words.

Every line is **cited** (`path:line`), **attested** with a measurement date, or a **dated incident** naming its commit. Anything genuinely unknowable sits in `Unverified` rather than being guessed.

## Review found real defects; all are fixed here

`Code Reviewer` approved; `Reality Checker` returned fix-then-ship. Both ran against `811a046`, and `3197038` closes everything.

**A security bug I introduced, confirmed live before fixing.** Filenames are PR-author controlled and a backtick is legal in a git path, so `` - `${n}` `` broke out of its code span:

```
.github/workflows/x`**INJECTED**`.yml   →   renders **INJECTED** as bold
```

No code execution — there is no shell step — but it injects markdown into a bot comment, and automation here reads PR comments as input. Fixed by whitelisting the path charset, which also covers newlines and raw HTML.

[...]

Issue #311, opened 2026-08-26
docs: CLAUDE.md and WORKFLOW.md misdescribe the CI gates — three verified inaccuracies, and a fuller audit is warranted


skizzy1986:
Found incidentally while building `.claude/skills/steward/SKILL.md`, which cites these files. All three are verified against the tree at `21370e8`. Every agent session reads `CLAUDE.md` as authoritative, so an inaccuracy here is not cosmetic — it is a wrong premise handed to every future session.

## 1. The gate count contradicts its own table

`CLAUDE.md:328`:

> Every PR is gated by **three** GitHub Actions jobs that must pass before merge:

The table immediately below has **four** rows: `Lint & Format`, `Test & Coverage`, `Build`, `Zone bands`. And `Zone bands` is not a `ci-web.yml` job at all — it is a separate workflow (`.github/workflows/zone-bands.yml`) with its own path filter.

`WORKFLOW.md:77` says *"Three required checks on every PR (`.github/workflows/ci-web.yml`)"*, which is right about `ci-web.yml` owning three but omits everything else that reports.

**Nine workflows actually report on a `web/**` PR**: `ci-web`, `ci-functions`, `ci-android`, `codeql`, `lighthouse`, `pr-title-lint`, `dependency-review`, `e2e-web`, `zone-bands`. Five appear in neither document.

**Suggested wording, `CLAUDE.md:328`:**

> Every PR is gated by the checks below. Three — `Lint & Format`, `Test & Coverage`, `Build` — are `ci-web.yml` jobs; the rest are separate workflows with their own path filters. Several more (`codeql`, `lighthouse`, `pr-title-lint`, `dependency-review`, `e2e-web`) also report on every PR without being listed here.

## 2. The `Lint & Format` cell is wrong in three ways

`CLAUDE.md:332` and `WORKFLOW.md:81` both document:

> ESLint errors + Prettier formatting + `npm audit --audit-level=high`

The job actually runs **five** commands (`ci-web.yml:51-69`):

```
npm run lint
npm run format:check
npm run check:design-sync
npm run check:colours
npm audit --omit=dev --audit-level=high
```

So the cell omits `check:design-sync` and `check:colours` entirely, and drops the `--omit=dev` scope — which is deliberate and load-bearing. `ci-web.yml:64-68` records why:

> An unscoped gate deadlocked the repo for a month in 2026-08 — it failed every web PR, including the Dependabot PRs carrying the fixes, and no single bump could clear it.

[...]

PR #316, opened 2026-08-28
docs(steward): record what protects main, and make Zone bands requirable

### LINKED ISSUE #312: test: coach-chat and vitals-sync have no automated verification — a PR touching them goes fully green with nothing run
A PR that touches only `supabase/functions/coach-chat/` or `supabase/functions/vitals-sync/` can pass every check on the PR while **nothing has run against the change**. This is not a gap in coverage percentage — it is the absence of any gate at all, presented as green.

## The mechanism

Three things combine:

1. **`ci-functions.yml` tests only two of the four functions.** It runs `deno run` against `vitals-import/test.ts` and `vitals-import-api/test.ts`. Neither `coach-chat/` (183 lines) nor `vitals-sync/` (159 lines) has a `test.ts` at all.
2. **Deno files are outside the lint and format scope.** `web/package.json` defines `lint` as `eslint src/ scripts/` and `format` as `prettier --write src/ scripts/`, both rooted at `web/`. Nothing under `supabase/` is reached.
3. **A skipped job counts as passing.** `ci-web.yml:3-7` states it plainly:

   > a job skipped by an `if:` condition reports "skipped", which branch protection counts as passing

   A PR touching only `supabase/functions/**` makes `ci-web`'s `changes` filter report `web == false`, so `Lint & Format`, `Test & Coverage` and `Build` all skip — and therefore pass.

Net: green PR, zero verification.

## This has already cost something

The deployed `vitals-sync` had drifted from its committed source — a different set of column names and four missing Google-Health fields, live in production against a v3 in git. From the `6904c18` commit body:

> nothing in CI would ever have caught it

That drift was found by hand, not by a gate.

## Why it matters more now

`coach-chat` is the Coach tab's only server path, and #301 showed how its failures present: a bare `TypeError: Failed to fetch` on the client with only the OPTIONS request in the edge logs. Silent on the dashboard, silent in CI.

`.claude/skills/steward/SKILL.md` currently carries a line warning agents not to trust green on these two functions, marked *provisional*. **That line should be deleted when this issue closes** — documenting a hole is strictly weaker than closing it, and the note exists only because the hole does.

## What would close it

Roughly in order of value per effort:

1. **A `test.ts` for `coach-chat` and `vitals-sync`**, matching the existing `vitals-import/test.ts` pattern — a pure, separately-testable parser or mapper exercised without network. `coach-chat` is a thin Anthropic proxy since #199, so its testable surface is small: request shaping, error mapping, the CORS header set.
2. **Wire those into `ci-functions.yml`** alongside the two that already run, so the job fails rather than passing vacuously.
3. **Extend lint/format to `supabase/`**, or add a Deno `fmt`/`lint` step — currently neither runs on any edge function.
4. **A drift check** comparing deployed function source against the repo, which is the specific failure `6904c18` describes. This is the one that would have caught the `vitals-sync` divergence, and is tracked separately at #210 / #304.

Items 1–2 are the minimum that makes green mean something.

## Acceptance

- [ ] `coach-chat` and `vitals-sync` each have a `test.ts` covering their pure logic
- [ ] `ci-functions.yml` runs both, and fails when they fail
- [ ] A deliberately broken `coach-chat` mapper turns the Edge Functions check red
- [ ] Deno files are covered by a format or lint step, or a comment records why not
- [ ] The provisional line in `.claude/skills/steward/SKILL.md` is deleted

skizzy1986:
Closes the `Unverified` slot #315 shipped with, and converts `zone-bands.yml` so the check it describes can actually be required.

**Scope note:** the conversion was originally deferred to a follow-up. It belongs here because it *invalidates a line this PR adds* — the skill would otherwise ship saying `Zone bands` cannot be required, and be wrong the moment the second PR merged. Landing them together keeps the file true at every commit.

## What protects `main`

Protection is a **Rulesets** entry named `main`, not classic branch protection — that page reads *"Classic branch protections have not been configured"*, which alone looks like `main` is unprotected. Active against the default branch with an **empty bypass list**, so it binds the repo owner too.

**Required — seven:** `Lint & Format` · `Test & Coverage` · `Build` · `CodeQL` · `Validate PR title` · `Review dependency changes` · `Deno Tests`

**Advisory:** `Zone bands` · `Lighthouse` · `Playwright Smoke` · `Playwright Visual` · `Build Debug APK`

Also on: a PR before merging, branches up to date, force pushes blocked **on `main` only**. **Linear history is not enforced** — squash is habit, not gate.

## The `zone-bands.yml` conversion

It used workflow-level `on.pull_request.paths`, so on a PR touching none of its inputs it reported **nothing** — not `skipped`. A required check that never reports leaves a PR on *"Expected — waiting for status"* indefinitely.

The failure would have been **intermittent, not obvious**: a PR carrying any `.md` passes, the next code-only PR hangs. The cause wouldn't have been visible from the PR that broke.

Now uses the `changes`-job pattern `ci-web.yml` and `ci-functions.yml` already use — path list moved verbatim into a `dorny/paths-filter` job, `check` gains `if: needs.changes.outputs.zones == 'true'`. Unmatched PRs report `skipped`, which branch protection counts as passing.

**The job name is deliberately unchanged.** `Zone bands match derivePaceZones` is the string that goes in the ruleset; renaming it would silently detach any requirement added later.

Concurrency also moves from `github.ref` to `head_ref || ref_name`, since `refs/pull/N/merge` and `refs/heads/<branch>` are different keys and wouldn't collapse if a push trigger is ever added.

`ci-android.yml` still has the old shape and is left alone — `Build Debug APK` remains unrequirable until it gets the same treatment.

## A nuance that looks like a fix and is not

`Deno Tests` is now required. But `ci-functions.yml` tests only the two `vitals-import*` functions, so on a `coach-chat` change it **runs, passes, and verifies nothing**. Requiring it did **not** close [#312](https://github.com/skizzy1986/erg-dashboard/issues/312).

Observed on this PR's own first run: **four of seven required checks were satisfied by `skipped`**. All correct — no web code, no edge functions — but it demonstrates that adding required checks raises the ceiling of what *can* block a merge, not the floor of what actually runs.

## The staleness rule fired inside the hour

`eb65fb9` recorded *"those three, any source. Nothing else."* True when written; wrong twelve minutes later when the ruleset was expanded. `5c19c75` corrects it. The mechanism working, and a fair warning about how fast that section rots.

One claim shipped in #315 becomes true rather than needing a fix: the skill says `ci-android.yml` *"alone"* uses workflow-level paths. That was **wrong when written** — `zone-bands.yml` did too. It is accurate now.

## A correction to #311

I filed [#311](https://github.com/skizzy1986/erg-dashboard/issues/311) claiming `CLAUDE.md:328`'s "three jobs" prose contradicted its four-row table, and proposed rewording **the prose**. Backwards: the prose was right; the table is wrong to list `Zone bands` as a gate. My proposed wording would have made the doc worse. Correcting separately on the issue.

## How to test manually

```bash
# the check still passes, script untouched
node web/scripts/check-zone-bands.mjs

# no workflow-level pull_request paths remain anywhere but ci-android
grep -A3 'pull_request:' .github/workflows/*.yml | grep -B1 paths
```

On this PR `Zone bands` should **run** (the diff has `.md`). The conversion's real proof is the next code-only PR, where it should report `skipped` rather than vanishing.

## Verification

- 990 tests · lint · `format:check` · `check:zones` clean; zone script passes over 286 markdown files
- `zone-bands.yml` parses; job name byte-identical to before; filter list moved verbatim
- `ci-android.yml` confirmed the only remaining workflow-level-`paths` user

## Checklist

- [ ] CI is green
- [x] Tests cover the change, or I've noted why they don't — docs plus a workflow trigger shape; the behaviour change is only observable in CI
- [x] Visual change: none
- [x] Stays inside the property boundary: no colour hex, box metric or type property

## Known gaps

- **`ci-android.yml` not converted** — same one-line-shape change if you want `Build Debug APK` requirable.
- **`Zone bands` is still only advisory.** This makes it *safe* to require; adding it is a ruleset edit I can't make.
- **The protection section rots fastest of anything in the skill** — it describes settings no file in the repo carries.

https://claude.ai/code/session_01ELg8XSCMQDcQswH6Him6Af

### DIFF
--- a/.claude/skills/steward/SKILL.md
+++ b/.claude/skills/steward/SKILL.md
@@ -13,7 +13,7 @@ or close/reopen a PR to kick CI; never approve or merge.
 
 Every line is **cited** (`path:line`), **attested** (`— measured YYYY-MM-DD`), or a **dated
 incident** naming the commit or PR that records it. Anything genuinely unknowable is
-quarantined under *Unverified* and stated nowhere else. A claim in none of those forms is a
+quarantined and marked as such rather than stated flat. A claim in none of those forms is a
 defect in this file. A line belongs only if an agent would act wrongly without it; general
 PR etiquette does not.
 
@@ -64,16 +64,26 @@ Green is at its most misleading here.
 
 1. Report what **ran**, not what was green. Name any job that reported `skipped`.
 2. If a fact here is contradicted by the file it cites, **say so, and offer to open an issue** — do not work around it silently. You are this file's staleness sensor.
-3. Never state a conclusion that depends on an unfilled slot below. Name the missing answer and what each answer would change.
+3. Name a **required** check by name when you say a PR is blocked. Three are required and the rest are advisory — see *What protects `main`* — so "CI is red" and "this cannot merge" are different claims.
 
-## Unverified — needs Scott
+## What protects `main`
 
-Do not guess, and do not default to the strict or the loose reading; both are wrong in a
-way that looks reasonable. Act up to where the answer changes the action, then ask.
+Read from the `main` ruleset — Rulesets, **not** classic branch protection, which is
+unconfigured (— confirmed by Scott 2026-08-28). Active, targeting the default branch, with
+an **empty bypass list**, so it applies to everyone including the repo owner. No agent can
+read this: the GitHub MCP server exposes no branch-protection tool and raw API access is
+blocked from agent sessions. Re-confirm with Scott rather than inferring it from a doc.
 
-| Question | What the repo already says | What is genuinely open |
-|---|---|---|
-| Which checks are **required** on `main`? | Four in-repo sources agree on **lint, test, build**: `ci-web.yml:4-5` ("Lint/Test/Build are REQUIRED status checks on main"), `dependabot-auto-merge.yml:3-4`, `CLAUDE.md:328`, `WORKFLOW.md:77`. Treat those three as required. | Whether protection has *since* added `Zone bands`, `E2E`, `Lighthouse`, `CodeQL` or `PR title lint`. No agent can read this: the GitHub MCP server exposes no branch-protection tool and raw API access is blocked. Until Scott confirms, **treat a red check from those five as blocking** rather than assuming it is advisory. |
+| | |
+|---|---|
+| Required status checks | **`Lint & Format`, `Test & Coverage`, `Build`** — those three, any source. Nothing else. |
+| Advisory — red does **not** block merge | `Zone bands`, `CodeQL`, `Lighthouse`, `Validate PR title`, `E2E`, `dependency review`, `Build Debug APK`. Still fix them; just do not report them as blocking. |
+| Also enforced | a PR before merging · branches up to date before merging · force pushes blocked **on `main` only**, so `--force-with-lease` on your own feature branch is fine |
+| Not enforced | linear history. Squash with `(#N)` is convention, not a gate. |
+
+**A required check can be satisfied by `skipped`** (`ci-web.yml:3-7`), so all three can be
+green having run nothing — see *When every check is green*. `CLAUDE.md:328`'s "three jobs"
+is correct; the four-row table under it wrongly lists `Zone bands` as a gate.
 
 ## Verified against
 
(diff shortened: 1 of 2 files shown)


skizzy1986:
Internal change — no release note. This PR only touches repo docs (`.claude/skills/steward/SKILL.md`) and CI workflow trigger mechanics (`.github/workflows/zone-bands.yml`); no user-visible behavior changed.

---

Issue #54, opened 2026-06-29
Strava OAuth2 activity import → sessions table


skizzy1986:
Roadmap integration, dormant 12+ months (schema-ready, no code). Sync Strava activities into the `sessions` table.

## Scope
- [x] Research the current Strava API (OAuth2 flow, rate limits, activity/stream endpoints) via `/research`. ✅ 2026-06-30 — see research comment below.
- [ ] OAuth2 connect flow; store tokens securely (never hardcode).
- [ ] Map Strava activities → `sessions` rows (type, label, duration, watts, hr, distance).
- [ ] Backfill + incremental sync; idempotent upserts.

The Strava MCP server is connected this session and can validate API calls during research. Spec via `/orchestrate` before building.

---
**Research decisions (2026-06-30):** full in-app OAuth2 connect flow · per-activity detail fetch for HR+watts · daily cron + on-demand sync. **Prerequisite:** Strava subscription required for API access as of 2026-06-30 (new Standard developer tier).

skizzy1986:
## Strava API research — build-ready findings

Research pass for the OAuth2 → `sessions` import. `supabase/functions/vitals-import-api/` is a near-perfect template. **TL;DR: build a polling Edge Function cloned from the vitals importer; one schema migration needed; refresh-token rotation is the one design subtlety.**

### ⚠️ Two schema discoveries (affect the spec)
1. **No `external_id` column** on `sessions` (it has `source`, but no per-source key). Need it for idempotent upsert. → add `external_id text` + partial unique index `UNIQUE (source, external_id) WHERE external_id IS NOT NULL`, upsert `ON CONFLICT (source, external_id)`. Strava's numeric activity `id` is the stable key.
2. **Live columns are `avg_watts` / `distance_m`** (renamed by migration `005`), not `watts`/`distance` as the brief/CLAUDE.md lists. `source` already exists.

### OAuth2 (confidence: HIGH)
- Authorize → `https://www.strava.com/oauth/authorize` (scope `activity:read_all`); exchange + refresh at `POST https://www.strava.com/oauth/token`.
- Access token **6 h**; refresh token has no time-expiry.
- Single-athlete flow = exactly the Google pattern: one-time gitignored local grant script (mirror `scripts/google-health-oauth.ts`) → seed tokens → Edge Function refreshes each run.

### Scope (HIGH)
- **`activity:read_all`** (plain `activity:read` omits private activities; for a personal dashboard you want everything; also covers reading own streams).

[...]

PR #323, opened 2026-08-28
fix(security): revoke anon/authenticated EXECUTE on SECURITY DEFINER functions

skizzy1986:
Found during a security review while working on #54 (Strava import).

## The hole

Supabase ships `ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT ALL ON FUNCTIONS TO anon, authenticated, service_role`. It fires at `CREATE FUNCTION` time, through the `PUBLIC` pseudo-role. No migration here ever revoked it, so **all six `SECURITY DEFINER` functions in `public` were `EXECUTE`-able by `anon`** — and `VITE_SUPABASE_ANON_KEY` is compiled into the shipped JS bundle by design.

`SECURITY DEFINER` is owned by `postgres`, so these functions **bypass RLS**.

| function | impact |
|---|---|
| `upsert_vital` | takes `p_user_id` as a **parameter** and writes to `vitals` with RLS bypassed — an anonymous caller can forge vitals rows for any `user_id` |
| `run_backup_snapshot` / `backup_snapshot` | invokable anonymously in a loop → full-DB JSON snapshots → storage/cost exhaustion |
| `fn_sync_strength_session`, `trg_strength_set_sync`, `trg_strength_workout_sync` | trigger/helper functions; revoked for consistency |

**Not a disclosure issue.** `backup_snapshots` has RLS enabled, zero policies, and no `select` privilege for `anon` or `authenticated`, so snapshot *contents* were never readable from the browser. Severity is data-integrity and resource-abuse, not a breach.

## Why signatures are resolved at runtime instead of hardcoded

This repo's migrations are a **mirror of a DB that has drifted** (same reason `007` exists), and two of the six targets cannot be named literally:

- `004` defines `upsert_vital` with **7** parameters, but [`vitals-sync/index.ts:124`](supabase/functions/vitals-sync/index.ts) calls it with **11** — `p_steps`, `p_distance`, `p_active_min`, `p_calories` were added live for the Google-Health columns and never mirrored back.
- `backup_snapshot` and `run_backup_snapshot` appear in **no migration at all**.

A literal `revoke ... on function public.upsert_vital(uuid,date,...)` would name a signature that may not exist and abort the whole migration. Looping `pg_proc` by name is correct against the drifted production DB *and* against a fresh DB rebuilt from `migrations/`, and no-ops on an absent function rather than failing.

## No caller breaks — checked, not assumed

- `web/src` contains **zero `.rpc(` call sites**. The browser's only call of this kind is `supabase.functions.invoke('vitals-sync')` ([`useVitalsSync.js:8`](web/src/hooks/useVitalsSync.js)), which runs inside the edge function.
- All **three** edge functions calling `upsert_vital` build their client with `SUPABASE_SERVICE_ROLE_KEY`, preserved by the `grant execute ... to service_role`: `vitals-import:95,100`, `vitals-import-api:119,120`, `vitals-sync:64,124`.
- The function **owner** (`postgres`) keeps EXECUTE regardless — `REVOKE ... FROM PUBLIC` does not touch owner privileges — so a `pg_cron` job scheduled as `postgres` is unaffected.
- Triggers fire as the table owner and **do not consult EXECUTE privilege**, so the strength rollup cannot break. Proven below.

## Verified against a real Postgres

I have no Supabase MCP connector in this session, so I could not apply this to `swdrueaserjzhuxnzmeu`. Instead I reproduced the Supabase role + default-privilege environment on `postgres:15-alpine`, including the **11-arg** drifted `upsert_vital`:

| # | check | result |
|---|---|---|
| 1 | vulnerable state reproduces | 6/6 functions `anon`+`authenticated` executable — matches the reported table exactly |
| 2 | migration clears it | 6/6 → `f` / `f`, `service_role` → `t` |
| 3 | `service_role` still writes vitals | row written ✅ |
| 4 | `anon` calls `upsert_vital` | `ERROR: permission denied for function upsert_vital` ✅ |
| 5 | `authenticated` calls `run_backup_snapshot` | `ERROR: permission denied` ✅ |
| 6 | trigger fires with no EXECUTE priv | `touched = t` ✅ |
| 7 | idempotent re-run | exit 0 |
| 8 | drift guard on an unknown secdef function | `ERROR: 011: SECURITY DEFINER functions still EXECUTE-able by anon/authenticated: some_future_fn()` ✅ |
| 9 | rollback restores exactly | back to 6/6 `t` |

## ⚠️ CI runs nothing against this diff

This PR touches only `supabase/migrations/**`. Every `ci-web` job gates on a `changes` job, so **`Lint & Format`, `Test & Coverage` and `Build` will report `skipped`** — which branch protection counts as passing. `Deno Tests` only covers the two `vitals-import*` functions. **Green here means "nothing ran", not "verified".** The Docker evidence above is the actual verification.

## Not done here, on purpose

The `ALTER DEFAULT PRIVILEGES` grant **is still in force**, so the *next* function created in `public` will again be born `anon`-callable. This migration is a one-time sweep, not a standing policy. Revoking that default is the real root-cause fix, but it changes the birth privileges of every future function in the schema and deserves its own reviewed change rather than riding along here. Worth a follow-up issue.

Also worth a follow-up: **`004` is stale** — it should be updated to the live 11-arg `upsert_vital`, and `backup_snapshot`/`run_backup_snapshot` should be mirrored into `migrations/` the way `007` mirrored the strength subsystem.

## Apply + verify (needs a Supabase connection)

1. Apply `011_revoke_security_definer_execute.sql` via `apply_migration` (named — not raw `execute_sql`).
2. Re-run the audit query; expect `f`/`f` on every row.
3. Confirm the vitals cron still lands a row — invoke `vitals-import-api` and read back.


skizzy1986:
Internal change — no release note. This locks down database-level EXECUTE privileges on `SECURITY DEFINER` functions (Supabase migration only); no user-visible behavior changes, and the PR confirms no legitimate caller relied on the closed access.

---

PR #325, opened 2026-08-28
ci(skills): guard the Context7-first rule on every workflow path

skizzy1986:
No issue — this came straight from a `/orchestrate` request. Happy to file one
retrospectively if the board wants the row.

## What changed and why

Context7 is connected in every session, but the *rule* to reach for it lived in
one skill's persona block — five of seven workflows never carried it, and that
copy had already drifted (no `@tanstack/react-query`, ESLint or Prettier). This
states the rule once and adds a guard so it cannot rot back out.

### The rule, stated once

`.claude/skills/erg-context.md` gains a canonical *Library documentation —
Context7 first* section. It is prepended verbatim to every pipeline spawn, so
`/orchestrate`, `/feature`, `/research` and `/refactor` inherit it without
repeating it. `/daily` and `/audit` spawn no erg-context agent, so each states it
where it bites — sizing a task, naming a root cause. `/orchestrate` additionally
makes it a `Code Reviewer` criterion: flag any library API the diff relies on
that was asserted from memory.

`/research`'s duplicated list is replaced by a pointer to the canonical block —
that duplicate is what drifted.

### The guard

`npm run check:context7` (`web/scripts/check-context7.mjs`) asserts every tracked
`SKILL.md` is reached by the rule via one of four routes:

```
check:context7: canonical block intact; 7 skill(s) reached (1 delegates, 2 direct, 3 inherits, 1 exempt).
```

Coverage is transitive, so the wording stays in exactly one place, and a new
skill carrying it by no route fails on arrival. `steward/SKILL.md` is exempt with
a written reason — its own contract requires every line to be cited, attested or
a dated incident, and says general PR etiquette does not belong. A **stale**
exemption fails too, the same discipline `check:colours` applies to its
allowlist.

### Why its own workflow, not a sixth step in `Lint & Format`

`ci-web.yml`'s `changes` job filters on `web/**` (`ci-web.yml:33-36`). A PR
touching only `.claude/skills/**` or the repo-root `CLAUDE.md` skips that job
entirely, so a guard living there would never fire on the files it exists to
protect. That is the same mistake `zone-bands.yml:10-14` was written to document.
Path filtering is at the job level, so it reports `skipped` rather than nothing
and **can be made required**.

### What it deliberately cannot do

It cannot verify an agent obeyed the rule — MCP calls are invisible to a CI
runner. It guards the *instruction*, not the *obedience*. That limit is stated in
the script header, the workflow comment and `CLAUDE.md`, so a green check is not
later mistaken for compliance.

## How to test manually

```bash
node web/scripts/check-context7.mjs
```

Sabotage-tested five ways — each exits 1 with a specific message, baseline exits 0:

| Sabotage | Caught |
|---|---|
| Canonical heading deleted from `erg-context.md` | yes |
| `daily` loses its direct mention | yes |
| A new uncovered skill lands | yes |
| An exempt file becomes covered (stale entry) | yes |
| `### Context7` removed from `CLAUDE.md` | yes |

## Follow-up for the repo owner

The check reports but does **not** block until it is added to the `main` ruleset.
The check-run name is **`Context7 rule reaches every workflow`**.

Read from the ruleset itself rather than from a doc
(`gh api repos/skizzy1986/erg-dashboard/rulesets/18063518`, `updated_at`
2026-06-24, `bypass_actors: []`), the required set is currently **three**:

```
Lint & Format · Test & Coverage · Build
```

So adding this one takes it from three to four. Note that `.claude/skills/steward/SKILL.md`
currently claims nine — that file is wrong, and separately from this PR the six it lists
(`CodeQL`, `Validate PR title (Conventional Commits)`, `Review dependency changes`,
`Deno Tests`, `Zone bands match derivePaceZones`, `Build Debug APK`) are still not
required. Worth its own issue; deliberately not folded in here.

## Checklist

- [x] CI is green (lint, test, build) — verified locally; see below
- [x] Tests cover the change, or I've noted why they don't — the guard is
      sabotage-tested above rather than unit-tested; it has no `src/**` surface
      and is outside the coverage scope (`include: ['src/**']`)
- [x] `npm run build` passes locally — bundle 395.7 KB / 400 KB, unchanged
- [x] No unexpected console errors in the browser — no runtime code touched
- [x] Visual change: none
- [x] Stays inside the property boundary — no colour, box-metric or type
      properties touched

Local gates: `lint` · `check:context7` · `check:zones` · `check:colours` ·
`check:design-sync` · **990 tests pass** · `build` · `size`. `format:check`
reports 182 files in this worktree, including untouched ones — that is a CRLF
checkout against Prettier's `endOfLine: lf`, not a real failure; the new script
was verified EOL-neutral against `check-zone-bands.mjs` as a control and is
Prettier-clean.

### DIFF
--- a/.claude/skills/audit/SKILL.md
+++ b/.claude/skills/audit/SKILL.md
@@ -69,6 +69,11 @@ Then **read the named file**. An issue is a report, not a diagnosis.
 - `analyze_issue_with_seer` is available for root-cause analysis. It is slow and
   may be metered, so run it only on findings that survive the read, and say when
   you are about to.
+- The **Context7 first** rule binds the verification itself. If the cause turns
+  on how a library behaves — a react-query cache eviction, a Recharts prop, a
+  `supabase-js` error shape — resolve it through `resolve-library-id` →
+  `query-docs` before naming it as the cause. A diagnosis from memory is the
+  same guess as a diagnosis from the Sentry title, one level down.
 
 Drop anything you cannot locate in the code. A finding you cannot point at is a
 guess.
--- a/.claude/skills/daily/SKILL.md
+++ b/.claude/skills/daily/SKILL.md
@@ -68,5 +68,9 @@ These are the difference between a useful brief and a harmful one.
 - Never invent an estimate. An unsized task says "needs a look first".
 - If `web/node_modules` is missing, state that `cd web && npm install` is step
   zero on every code task — don't assume it.
+- **Context7 first** whenever sizing a task turns on a library fact — whether a
+  Vitest option exists, what a react-query upgrade changes. Resolve it
+  (`resolve-library-id` → `query-docs`) or say "needs a look first". A
+  remembered API is exactly how a 30-minute task becomes a two-hour one.
 - Re-rank from scratch every run. Never carry a previous list forward.
 - Six items is the ceiling, not the target. Three good ones beat six padded.
--- a/.claude/skills/erg-context.md
+++ b/.claude/skills/erg-context.md
@@ -45,6 +45,25 @@ Testing Library. Full briefing: `CLAUDE.md`.
 - Read back every write. Destructive changes require Scott's explicit approval.
 - Deeper patterns: `.claude/skills/supabase-patterns.md`.
 
+## Library documentation — Context7 first
+
+The Context7 MCP connector is live in every session. **Never answer a library
+question from memory, and never reach for WebSearch first.** Look it up:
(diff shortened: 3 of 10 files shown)


skizzy1986:
Internal change — no release note. This PR adds a CI guard (`check:context7`) and consolidates Context7-lookup guidance in `.claude/skills/` and `CLAUDE.md`; it touches no application code and has no user-visible effect.


---

PR #328, opened 2026-08-28
fix(steward): correct the required set, and the guard bug that correcting it exposed

skizzy1986:
Follow-up to #325 (now merged). No issue — surfaced while stewarding that PR,
which is exactly what rule 2 of *When you report back* asks an agent to do.

> **Update after rebase onto merged #325.** This PR now carries a second commit:
> the Context7 guard flagged *this very change*, and it was right to look but
> wrong about why. Details in **Second commit** below.

## First commit — `docs(steward): correct the required set — read the ruleset API, not the page`

**The required-check set in `steward/SKILL.md` was wrong, and had been for three
PRs.** It claimed seven (#316), then nine (#321), each with a dated attestation
citing "the ruleset page as Scott showed it". Neither was ever true — the ruleset
had required exactly three checks since 2026-06-24 and had not been edited since:

```
GET /repos/skizzy1986/erg-dashboard/rulesets/18063518
  → ["Lint & Format", "Test & Coverage", "Build"]
  → updated_at 2026-06-24T16:35:12   (read twice, ten minutes apart)
```

The interesting part is **why it survived**. This same file said:

> No agent can read this: the GitHub MCP server exposes no branch-protection tool
> and raw API access is blocked from agent sessions.

That is false in a local session — `gh api …/rulesets/18063518` works — and it
told every agent not to look. So sessions corroborated each other's assumption
instead of checking the source. `CLAUDE.md`'s CI table was the only doc that
matched reality.

This commit:

1. **Replaces the unreadable-by-agents claim with the command that reads it**, and
   leads the section with *"read it yourself — do not trust the table below, this
   file included."*
2. **Records the incident.** A file whose entire contract is that every line is
   cited, attested, or a dated incident should carry its own worst failure — and
   the generalisable lesson is sharp: *when this file tells you a capability is
   unavailable, test the capability before believing it; that instruction is
   itself the thing most likely to be stale.*
3. **Corrects the set to ten.** Scott added the six missing checks plus
   `Context7 rule reaches every workflow` on 2026-08-28 at 12:22 +08:00 — which is
   when the table first became true. Also notes that `CodeQL` is emitted by the
   GitHub Advanced Security app and is a **different check-run** from `codeql.yml`'s
   job `Analyze (JavaScript/TypeScript)`; the ruleset binds the former.
4. **Re-attests against the API response** — `updated_at`, `bypass_actors`,
   `enforcement` — not a screenshot, and says to re-run rather than trust the
   timestamp: the ruleset is edited outside this repo, so nothing in git can tell
   you it moved.
5. Notes why `Context7 rule reaches every workflow` is its own workflow rather
   than a step in `Lint & Format`.

## Second commit — `fix(ci): match the Context7 rule by its lookup tool, not the bare word`

Correcting the required set necessarily made `steward/SKILL.md` mention
`Context7 rule reaches every workflow` — the **name of the required check-run**.
The guard's `direct` coverage route tested `text.includes('Context7')`, scored the
file as covered, collided that with its `EXEMPT` entry, and failed as a stale
exemption:

```
✗ .claude/skills/steward/SKILL.md is listed as exempt but is now covered (direct)
  — delete its EXEMPT entry in web/scripts/check-context7.mjs
```

**The exemption was not stale.** `steward` still does not carry the rule and
should not — it names a check, not a lookup. The defect was the predicate: a bare
substring on the word "Context7" is not coverage, it is a word.

`direct` is now keyed on the lookup tool. A skill carries the rule only if it
tells you what to actually call:

| skill | says `Context7` | says `resolve-library-id` | route |
|---|---|---|---|
| `daily`, `audit` | yes | yes | `direct` |
| `steward` | yes | **no** | `exempt`, correctly |

Routing is unchanged — `1 delegates, 2 direct, 3 inherits, 1 exempt` — and the
failure message now names the tool to add rather than the word. Re-sabotaged:
stripping `resolve-library-id` from `daily` still fails with

```
✗ … is not reached by the Context7-first rule. Prepend …/erg-context.md to its
  spawns, name `resolve-library-id` in the file itself, or add a reasoned EXEMPT entry
```

Worth saying plainly: the guard caught a real ambiguity in its own definition of
coverage on the second change it ever saw. That is the check working, not
misfiring.

## How to test manually

```bash
gh api repos/skizzy1986/erg-dashboard/rulesets/18063518 --jq '{updated_at, bypass: .bypass_actors, required: [.rules[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context]}'
```

Should return ten contexts, `bypass_actors: []`, `updated_at` 2026-08-28T12:22:08.
If it does not, this file is stale again — and the whole point of the change is
that you find that out by running the command rather than by reading the table.

```bash
node web/scripts/check-context7.mjs
```

## Checklist

- [x] Tests cover the change, or I've noted why they don't — the doc claims are
      verified against the live ruleset API, quoted above; the guard change is
      sabotage-tested rather than unit-tested and is outside the coverage scope
      (`include: ['src/**']`)
- [x] `npm run build` passes locally — pre-push hook ran the full Vitest suite
- [x] No unexpected console errors in the browser — no runtime code touched
- [x] Visual change: none
- [x] Stays inside the property boundary — no colour, box-metric or type
      properties touched

### DIFF
--- a/.claude/skills/steward/SKILL.md
+++ b/.claude/skills/steward/SKILL.md
@@ -64,26 +64,46 @@ Green is at its most misleading here.
 
 1. Report what **ran**, not what was green. Name any job that reported `skipped`.
 2. If a fact here is contradicted by the file it cites, **say so, and offer to open an issue** — do not work around it silently. You are this file's staleness sensor.
-3. Name a **required** check by name when you say a PR is blocked. Nine are required — see *What protects `main`* — so "CI is red" and "this cannot merge" are different claims. Use the **check-run name**, which is the job's `name:`, not the workflow's: the ruleset binds to `Zone bands match derivePaceZones`, and nothing reports under `Zone bands`.
+3. Name a **required** check by name when you say a PR is blocked. Ten are required — see *What protects `main`*, and read the ruleset rather than trusting that number — so "CI is red" and "this cannot merge" are different claims. Use the **check-run name**, which is the job's `name:`, not the workflow's: the ruleset binds to `Zone bands match derivePaceZones`, and nothing reports under `Zone bands`.
 
 ## What protects `main`
 
-Read from the `main` ruleset — Rulesets, **not** classic branch protection, which is
-unconfigured (— confirmed by Scott 2026-08-28 from the ruleset page itself; required set expanded
-from seven to nine later the same day). Active, targeting the default branch, with
-an **empty bypass list**, so it applies to everyone including the repo owner. No agent can
-read this: the GitHub MCP server exposes no branch-protection tool and raw API access is
-blocked from agent sessions. Re-confirm with Scott rather than inferring it from a doc.
+**Read it yourself — do not trust the table below, this file included.**
+
+```bash
+gh api repos/skizzy1986/erg-dashboard/rulesets/18063518 \
+  --jq '{updated_at, bypass: .bypass_actors, required: [.rules[]
+    | select(.type=="required_status_checks")
+    | .parameters.required_status_checks[].context]}'
+```
+
+Rulesets, **not** classic branch protection: `GET /branches/main/protection` returns 404.
+Active, targeting the default branch, `bypass_actors: []`, so it applies to everyone
+including the repo owner.
+
+**This section was wrong for three PRs, and the reason it survived is a warning about
+the whole file.** It claimed seven required checks (#316), then nine (#321) — each with a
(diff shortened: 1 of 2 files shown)


skizzy1986:
Internal change — no release note. This PR corrects internal agent-pipeline documentation (the required GitHub status-check set in `steward/SKILL.md`) and fixes a matching predicate bug in the `check-context7.mjs` CI guard. No application code, UI, or end-user-facing behavior is affected.

---

PR #319, opened 2026-08-28
docs: correct the bundle headroom figure — 4.3 KB, not ~40 KB

skizzy1986:
Related to #311 (the "check the fast-moving sections claim-by-claim" item), but fixed directly rather than folded in — it is a single verified number, not part of the CI-gate rewrite that issue is about.

## What changed and why

`CLAUDE.md:296` claimed **~40 KB** of headroom under the 400 KB gzip budget. The real figure is **4.3 KB** — stale by an order of magnitude, and stale in the direction that invites someone to spend room that is not there.

Re-measured on this branch via `npm run size`:

```
    391.7 KB  index-B7Lm0UPw.js
      1.5 KB  web-B-UxAMWU.js
      1.3 KB  web-D0ZZATOk.js
      0.5 KB  web-CVoKHV0s.js
      0.4 KB  web-DSxpgqtr.js
      0.3 KB  index-AgTDnBZY.css
  ──────────
    395.7 KB  total  (budget 400 KB)

✓ Within budget (4.3 KB headroom).
```

Unchanged from the 2026-08-26 reading already attested in the steward skill, so the two now agree on a measured number rather than one contradicting the other.

## Why it is not just a number swap

The sentence was making an argument, and the wrong number weakened it. Session Replay at ~35–50 KB read as a close call against ~40 KB of room — the kind of trade-off someone might reasonably revisit. Against 4.3 KB it is not close, and the actual blocker is structural: nothing in `App.jsx` is lazy-loaded, so every import ships on first paint. Enabling Replay is a code-splitting job, not a config flag. The rewrite says that, so the next person to consider it starts from the real constraint.

## Second file: the staleness sensor was itself stale

`.claude/skills/steward/SKILL.md:53` carried the clause that flagged this very defect:

> `CLAUDE.md:297` still claims ~40 KB; it is stale by an order of magnitude.

Once `CLAUDE.md` is fixed, that clause describes a state that no longer exists. Leaving it would make the file that exists to catch stale claims the stale claim — and its own rule 2 says a fact contradicted by the file it cites is a defect in that file. Updated to record that the two now agree, with the re-measurement date.

## How to test manually

```bash
cd web && npm run build && npm run size
```

Confirm the reported total and headroom match the figures now in `CLAUDE.md`.

## Verification

Ran the full `Lint & Format` gate — all five commands, not just `lint`:

| Check | Result |
|---|---|
| `lint` | clean |
| `format:check` | all matched files use Prettier style |
[...]

PR #329, opened 2026-08-28
perf(web): drop mathjs, reclaiming 52 KB of bundle budget

skizzy1986:
## Problem

The bundle sat at **395.7 KB against the 400 KB gate — 4.3 KB spare**, less than the next feature PR (#54, Strava import, ~3.1 KB) needed. Two contributing issues: the documented figure was stale (`~360 KB` with `~40 KB` headroom), and nobody had looked at where the bytes went.

## What this does

**`mathjs` was 52 KB of the bundle**, imported from the package root to supply `mean` and `std` on plain number arrays, inside one tab's regression panel. The least-squares fit sitting beside them was always hand-rolled JS — mathjs has no `linearRegression` function and never did, despite what `training-science.md` claimed.

Replaced with `web/src/utils/stats.js` (+ 8 tests).

> `sampleStdDev` deliberately preserves mathjs's default `'unbiased'` n−1 divisor. This is not cosmetic: on the real HR130 anchors `[145, 151, 149, 149]` the rendered spread is **2.5**, where a population `n` divisor would silently print **2.2**. A test pins the exact mathjs value.

## Measured result

| | total gzipped | headroom |
|---|---|---|
| before (`origin/main`) | 395.7 KB | 4.3 KB |
| **after** | **343.7 KB** | **56.3 KB** |

## Why not code-splitting

The task suggested route/tab-level splitting as "likely the single biggest win". I measured it, and **for this gate it is a loss**.

`check-bundle-size.mjs` sums **every emitted asset** — total bytes shipped, not the initial chunk. Putting all 13 desktop tabs behind `React.lazy` + `Suspense`:

- entry chunk: 339.7 KB → **202.5 KB** (a genuine ~40% first-paint win)
- **total: 343.7 KB → 364.4 KB (+20.7 KB)** in chunk overhead and duplicated shared code

So it would have spent 20.7 KB of the very headroom this PR set out to reclaim. Splitting is still worth doing on its own terms, but it belongs with a change to *what the gate measures* (entry chunk vs. total) — left as a follow-up rather than smuggled in here. The reasoning is now recorded in `check-bundle-size.mjs` so the next person doesn't reach for `React.lazy` to fix a red budget.

`MAX_GZIP_KB` is **unchanged at 400** — no rubber-stamping.

## Docs

- `CLAUDE.md` — real figures, the two load-bearing properties of the gate, and a standing **"never quote a bundle figure from this file — re-measure it"** instruction, since these have already drifted once (#319). Also drops mathjs from the tech-stack table.
- The Session Replay paragraph said there was "nothing like that much room". With 56.3 KB free, Replay's ~35–50 KB now fits — so that is recorded as a live choice to re-make, not a byte constraint.
- `.claude/skills/steward/SKILL.md` — updated figure + the anti-pattern.
- `.claude/skills/training-science.md` — corrects the `mathjs linearRegression` claim.
- `check-bundle-size.mjs` — its own comment still said "~360 KB"; now points at the generated `dist/bundle-size.json` instead of naming a number that can rot.

## Verification

- `npm test` — **998 passed** (79 files), incl. 8 new
- coverage — lines 89.06 / functions 83.44 / branches 80.74, all above the 85 / 79 / 76 ratchet
- `npm run lint` — clean
- `npm run build` — clean; `npm run size` — 343.7 KB, ✓ within budget
- `check:zones`, `check:colours`, `check:design-sync`, `npm audit --audit-level=high` — all pass

<sub>Note: `npm run format:check` flags ~179 files on a Windows working tree — `core.autocrlf=true` with no `.gitattributes`, and Prettier defaults to `endOfLine: lf`. Git stores LF, so CI is unaffected. Deliberately not "fixed" here, as reformatting 179 files would bury this diff.</sub>


skizzy1986:
Merged `main` in to clear the conflict #335 introduced, and re-measured. **All checks green.**

**Conflict resolution.** Both files conflicted on the same thing — #335 bumped three
dependencies this branch's `package.json` also touches. Took main's bumps
(`@sentry/react` 10.71.0, `@supabase/supabase-js` 2.112.4, `@tanstack/react-query`
5.102.8) and kept mathjs removed, which is the point of the PR. The lockfile was
regenerated from main's against the resolved manifest rather than hand-merged.

**One addition:** the branch also carries #340's `browserslist`/`fast-uri` bump. Without
it `Lint & Format` fails here on `npm audit`, an advisory that is red on `main` and has
nothing to do with this PR (#339). It no-ops once #340 lands.

**The table in the description is now slightly stale** — `main` has grown since it was
written. Re-measured on the merge commit:

| | total gzipped | headroom |
|---|---|---|
| `main` today | 399.1 KB | 0.9 KB |
| **this branch** | **347.2 KB** | **52.8 KB** |

The reclaimed amount is unchanged at ~52 KB; both endpoints just moved up ~3.5 KB. The
"0.9 KB spare on main" figure is worth noting on its own — that is tighter than the
4.3 KB the description cites, so this PR is more load-bearing than when it was opened.

Verified on the merge commit: 1078 tests pass, lint/format/build clean, audit clean, and
`check:zones` / `check:colours` / `check:design-sync` / `check:context7` all pass.

---

skizzy1986:
Internal change — no release note. This swaps the `mathjs` dependency for a hand-rolled `stats.js` utility to cut bundle size (395.7 KB → 343.7 KB), with docs and tests updated to match. Output values are unchanged (a test pins the exact prior `mathjs` result), so there's no user-visible behavior change.

---

PR #334, opened 2026-08-28
perf(web): budget the entry chunk, then split the shells and tabs

### LINKED ISSUE #331: Bundle gate measures total bytes, so it penalises the code-splitting it asks for
## What

`web/scripts/check-bundle-size.mjs` sums **every** emitted asset in `dist/assets`
and gates that total against `MAX_GZIP_KB = 400`. Total bytes shipped is not what
a first paint costs — the entry chunk is. Because splitting trades a smaller entry
chunk for a slightly larger total (chunk preamble + shared code that lands in more
than one chunk), the gate reports a genuine user-facing win as a regression.

Measured on the PR #329 branch (post-mathjs, total 343.7 KB): putting the 13 desktop
tabs behind `React.lazy` cut the entry chunk **339.7 KB → 202.5 KB** (~40% off first
paint) while total went **343.7 → 364.4 KB** (+20.7 KB). Under today's gate that is
a failure.

The doc trail says the same thing from the other side — `CLAUDE.md:296-301` and
`.claude/skills/steward/SKILL.md:53` both record "nothing in `App.jsx` is lazy-loaded,
so every import ships", with 4.3 KB of headroom against a total-bytes budget.

## Why now

Session Replay (~35-50 KB gzip) and any future feature of size are blocked behind
this. The gate currently makes the fix illegal.

## Acceptance criteria

- [ ] `check-bundle-size.mjs` gates the **entry chunk** (entry + its transitive
      *static* imports + their CSS), identified from the Vite/rolldown manifest —
      not guessed from filenames
- [ ] A separate, looser **total** budget still guards against unbounded growth
      (a heavy dep on a lazy route must not be free)
- [ ] `dist/bundle-size.json` keeps `budgetKb` / `total` / `assets` so
      `report-bundle-size.mjs` still diffs against an old-shape main baseline
- [ ] The PR comment reports entry delta and total delta
- [ ] `AuthGate.jsx` no longer statically imports both `App.jsx` and
      `views/mobile/MobileApp.jsx` — a device renders one shell, not two
- [ ] The 13 desktop tabs in `App.jsx` are lazy, behind the **existing**
      per-tab `ErrorBoundary` (which reports via `componentDidCatch` to Sentry)
- [ ] Suspense fallbacks do not flash on tab switches
- [ ] Both `CLAUDE.md` and `.claude/skills/steward/SKILL.md` are updated to the
      new semantics and re-measured figures

Follow-up to #329.

skizzy1986:
Closes #331. Follow-up to #329. Rebased onto #324 (Strava), which added a 14th tab.

## The problem

`check-bundle-size.mjs` summed **every** emitted asset in `dist/assets` and gated that
total at 400 KB. Total bytes shipped is not what a first paint costs, and splitting
always trades a much smaller entry chunk for a slightly larger total — per-chunk
preamble, plus shared code that now lands in more than one chunk. So the gate reported
the code-splitting it kept asking for as a regression.

**This was not a theoretical bind.** `main` is at **399.2 KB of its 400 KB budget —
0.8 KB spare** ([Build log](https://github.com/skizzy1986/erg-dashboard/actions/runs/33147274731)).
The next commit of any size would have failed it, and the fix the message suggests
("code-split") would have failed it harder.

## 1. Change what the gate measures

| | What | Budget | Measured |
|---|---|---:|---:|
| **Entry** | entry chunk + its transitive **static** imports + their CSS | 150 KB | **134.7 KB** |
| **Total** | every emitted asset, lazy ones included | 450 KB | **425.5 KB** |

The entry set is read from `dist/.vite/manifest.json` (`build.manifest` is now on in
`vite.config.js`) — the only source that distinguishes a static import from a dynamic
one. Filenames cannot, and guessing would make the budget a fiction.

The walk lives in `scripts/entry-graph.mjs` with its own tests, because **its failure
mode is silent**: follow `dynamicImports` by mistake and "entry" quietly becomes the
total again — every chunk reachable, budget passing, gate meaningless. A wrong number
there does not throw. The tests pin exclusion of lazy chunks, transitive static
following, the cycle (lazy chunks statically import back into the entry record), and
that a manifest with no `isEntry` throws rather than reporting 0 KB.

`dist/bundle-size.json` keeps `budgetKb` / `total` / `assets` with their original
meanings, so the baseline `main` has cached today still reads; `entry`,
`entryBudgetKb` and `entryAssets` are added alongside. `report-bundle-size.mjs` was
checked against all three baseline shapes (old-shape, new-shape, absent) and says
explicitly when a baseline predates the entry budget rather than inventing a delta.

## 2. Land the split

- `AuthGate.jsx` statically imported **both** `App.jsx` and `MobileApp.jsx`, so every
  phone downloaded a 14-tab dashboard it cannot reach and every desktop downloaded a
  phone app it cannot reach. Both are lazy.
- All 14 tabs in `App.jsx` are lazy.

```
entry  395.4 → 134.7 KB   (−66%)
total  399.2 → 425.5 KB   (+26.3 KB — expected, and now budgeted separately)
```

**The existing per-tab `ErrorBoundary` is preserved and still reports.** `Suspense`
sits *inside* it, so a chunk that fails to load or a tab that throws on first render
is still caught and still reaches Sentry via `componentDidCatch`. The existing
`App.test.jsx` error-boundary test is what pins that arrangement.

**Fallbacks don't flash.** Tabs use `fallback={null}` — a tab is a full-page swap, and
`React.lazy` caches each module, so only a tab's *first* visit can suspend at all. The
shell boundary in `AuthGate` reuses the existing `Loading…` pane (lifted to
`LoadingPane` so the two cannot drift), and renders `null` while the mobile splash is
already covering the window — so a slow shell chunk reads as loading, not as a blank
screen.

## Verification

- `npx vitest run --coverage` — **1076 passed**, coverage 89.81 lines / 84.70 functions
  / 81.51 branches, above the 85/79/76 ratchet.
- `npm run lint`, `check:design-sync`, `check:colours`, `check:zones`, `check:context7` — clean.
- Prettier is clean on every file this PR touches, verified EOL-neutrally against an
  untouched control. (Raw `--check` flags CRLF in a Windows worktree; CI sees LF.)
- **Playwright, in the pinned `v1.62.1-noble` container**: smoke **14/14**, visual
  **10/10**, no baseline churn.
- `smoke.spec.js` now waits for the chunk before asserting. Both its checks are
  *absence*-checks, which pass trivially against a view that has not mounted — without
  the wait the suite would have gone green while proving nothing. **Mutation-tested**:
  with `JournalView` patched to throw, the journal smoke test fails as it should.

## Docs

`CLAUDE.md` and `.claude/skills/steward/SKILL.md` both documented the old single-budget
gate as the reason splitting was blocked. Both now carry the two-budget semantics, the
re-measured figures, and state plainly that **a PR raising `total` while cutting
`entry` is a win, not a regression**.

## Notes for review

- **One addition beyond the issue's scope:** the settings tab from #324 was in neither
  `smoke.spec.js`'s `TABS` nor `App.test.jsx`'s `NAV`. This PR makes it lazy, so it is
  added to both rather than left as the one lazy tab whose mount nothing checks.
- **`MAX_TOTAL_GZIP_KB = 450` is set against a `main` that still has mathjs.** #329
  removes ~52 KB. When it lands, ratchet total down rather than banking the slack —
  the comment at `check-bundle-size.mjs:31-35` says so.
- Session Replay still doesn't fit (~35-50 KB vs 15.3 KB entry headroom, and
  `Sentry.init` runs at first paint), but it's now a `lazyLoadIntegration` question
  rather than a flat impossibility.
- **recharts** (90.7 KB, 7 importers) needed no trimming — the split takes it out of
  first paint entirely.
- **supabaseClient** stays in the entry, correctly: auth resolves before anything can
  render. Worth its own issue: `createClient` pulls in `RealtimeClient` unconditionally
  and nothing in `src/` uses realtime, but reclaiming it means dropping to `auth-js` +
  `postgrest-js` directly.


skizzy1986:
#329 has landed, so this now does what `check-bundle-size.mjs:31-35` said to do when it did: **ratchet the total budget down rather than bank the slack.**

Re-measured on the merge:

| | Gzipped | Budget | Spare |
|---|---:|---:|---:|
| Entry | 134.9 KB | 150 KB | 15.1 KB |
| **Total** | **373.5 KB** | **400 KB** *(was 450)* | 26.5 KB |

Total fell 425.4 → 373.5 KB — the ~52 KB #329 predicted, near enough exactly. The new
ceiling preserves the ~25 KB of headroom the original note specified. Entry is unchanged.

### Conflicts

Merging `main` conflicted in three files, all because #329 rewrote the same passages.
In each case this branch's text is the successor — main describes a *single* total-bytes
budget, which stops being true once this PR lands — so HEAD was kept and main's
still-valid independent points were carried across rather than dropped:

- `check-bundle-size.mjs` — kept the two-budget explanation; carried over #329's
  "don't quote a size figure from memory, run `npm run size`" instruction.
- `CLAUDE.md` — same, plus the **"never quote a bundle figure from this file"** paragraph
  and its #319 history, which this branch did not have.
- `.claude/skills/steward/SKILL.md` — kept both of this branch's rows; carried over the
  re-measure caution that only main's row carried.

Figures in `CLAUDE.md` and the steward skill are updated to the re-measured numbers, so
the two documents still state the same thing — which the steward row explicitly claims.

> One correction worth recording: I initially resolved only two of the three conflicts,
> because the merge output was truncated in my terminal and I staged the third file
> explicitly. The markers were committed locally and caught by a full-tree scan before
> the push — `9bf5068` fixes it. Nothing with markers reached the remote.

### Verification

`npm test` — **1084 passed** (87 files) · lint, `format:check`, build clean ·
`check:zones` / `check:colours` / `check:design-sync` / `check:context7` all pass ·
`npm run size` within both budgets.

---

skizzy1986:
`Playwright Visual` went red on `desktop shell › program` — **164765 pixels, 17% of the image**. That is a real defect this PR introduces, not a flake, and it is now fixed in `ae6195b`.

### What it is

`switchTo()` polls for `#root` innerHTML to differ from before the click. That was sound
while every tab was statically imported. Now that all 14 sit behind `React.lazy` with
`fallback={null}` (`App.jsx:402`), the poll no longer proves the new tab rendered.

I measured `#root` innerHTML length across a PROGRAM switch on this branch:

```
before click  42989
poll satisfied 43219   ← switchTo() returns here, screenshot fires
chunk loaded  73883
```

The **~230 characters** that satisfy the poll are the nav button's own active styling.
The tab's ~30 KB of markup — 71% of the final tree — has not arrived. So the screenshot
captures the *outgoing* tab under a highlighted nav button, which is exactly a 17% diff.

### Why it passed on the previous head

It is a race, so it passes whenever the chunk wins. Identical app code went green on
`9bf5068` and red on `9d969a7`. Worth being explicit that this would not have stayed
hidden: it would have flickered on `main` after merge instead.

### The fix

`await page.waitForLoadState('networkidle')` after the switch — **the same remedy this PR
already applied to `smoke.spec.js`**, for the same reason. That file's note says its
absence-checks "would have gone green while proving nothing" against an unmounted view;
`visual.spec.js` has the identical race and was missed. Supabase is stubbed from the route
handler, so the chunk is the only request left in flight.

No baseline was regenerated. The baseline is correct — the screenshot was early.

> Credit where due: the `visual` project sets `retries: 0` with the comment *"a retry would
> paper over exactly the flake this harness exists to make visible."* That is the only
> reason this surfaced rather than passing on attempt two.

`lint` and `format:check` clean on the change.

---

skizzy1986:
**Release note draft — Improved**

The app now loads faster, especially on the first visit or right after a browser cache clear. Dashboard tabs, the strength logger, and the mobile view are now fetched only as you open them instead of all being downloaded up front, cutting the initial page weight by roughly two-thirds. No functional changes — everything works the same, it just gets there quicker.

---

